PR #8, opened 2024-09-07
Integrating streamlit app updates

katiefg:
# Description

Updated files in streamlit_app including app.py, the_map.py, requirements.txt within streamlit_app directory, and data files for experimentation with the streamlit app concept. Only files in streamlit_app are impacted by this merge.

## Type of changes
- [x] New Feature (exploratory streamlit app concept)

## Testing
- I did not add automated testing

## How to test
-Install requirements.txt
-From streamlit_app/archive/streamlit_app_concept, run streamlit run app.py
-Enter an address into the search bar, press "enter", see whether information is provided and accurate


oscarsyu (approved):
Really awesome work thank you for putting this together!  It's definitely better than anything I've put together in streamlit from a technical standpoint so its super cool to see how you went about designing this.

I think there's some valuable visualization ideas that you should share with the team, if you haven't done so already.  And i like the concepts for some of the analytics you've pulled.

Hopefully the team has filled you in on this, but the streamlit app was something I just threw together as a concept and won't be the actual thing we launch.  I do, however, think it's a valuable place to test out ideas and interactivity.  Feel free to keep developing it (and I can help too), or just experiment with visualizations and analytics through jupyter notebooks as well.

Here's maybe some feedback overall:
- a bit of a nit, but the show function is pretty long and ideally we'd want to break it into pieces.  but also, this should not be a prioritiy at all here

maybe more importantly:
- I think the visualizations are great, especially the animations and counts of soft story statuses.  
- It does look a little busy and I think it might be better if we can go with an approach where we can filter and plot addresses as dots since this craiglist style interface is probably better for searching from a neighborhood level but not great for searching on a address level.  
- the way the layerying on the map to me is a little busy and difficult to quickly read over it.  perhaps in the final mock up, we want to only enable 1 or 2 tsunami, or liquification zone layers at a time to avoid overplotting
- im wondering if the color coding approach you've taken that groups things up by green, orange, or yellow might be better done on a neighborhood level, eg: Duboce Triangle is all "green" instead of grouping into a number count.

overall though, take this feedback as suggestions and great work.  really impressive and appreciate hopping on and contributing so quickly!

oscarsyu:
i'll also mention, when i tried plotting everything on the map, i ran into major performance issues, so we would need a way to get around that.

Issue #35, opened 2024-09-17
Set Up Routing for Core Pages


mantuok:
## Context
The app will have multiple pages, we need to set up the routing, including dynamic routing for specific entities. The preliminary pages are: Homepage, Selected address, Resources, About, and Methodology.

## Definition of Done
1. The pages are functional
2. Dynamic routing is configured
3. Navigation works correctly
4. Fallback pages for invalid routes are implemented

## Engineering Details
Use the Next.js 13 App Router system and organize pages within the app directory.

_Testing Criteria_
- Each page displays a placeholder with the page name.
- Navigation between pages works without errors.
- Dynamic routes display the correct content for entities.
- Invalid routes show the fallback page (e.g., 404 error).

mackcooper1408:
I think we can actually begin working on this asap. The current Figma should have the first couple of pages laid out, so I think if you update this issue to specify exactly which pages we need and link to the figma, then you can mark this as ready to implement. I also think it's appropriate to add some testing criteria (i.e. for each page x, y, and z, a user should be able to see a "dummy" placeholder page that says the name of the page).

PR #48, opened 2024-09-26
Set up routing

mantuok:
# Description

Created main pages to support the routing functionality. Ensured proper linking between pages. Added not-found page for non-existing pages. Implemented a dedicated not-found page.

#35 

## Type of changes
- [ ] Bugfix
- [ ] Chore
- [x] New Feature

## Testing
- [ ] I added automated tests


app/mapView/page.tsx
```diff
@@ -0,0 +1,18 @@
+import { Heading, Text } from "@chakra-ui/react";
```
mackcooper1408, line 1: I don't feel _too_ strongly either way, but I'm leaning towards Kebab case and naming the file `/map-view/page.tsx` instead of `/mapView/page.tsx`. And that's what you do for th 404 page, so as long as we're consistent I'm happy 😄

mackcooper1408 (approved):
Looks great! One nit about file naming, otherwise good to merge whenever you want

app/mapView/page.tsx
mantuok: Done! :)

Issue #39, opened 2024-09-18
Build a Component Test Page for Showcasing UI Components


mantuok:
_Can be implemented once issue #38 is done_

## Context
Build a component test page that provides an accessible location for reviewing all the components built so far. This will help with visual inspection of the components, ensuring they behave correctly in various states. In addition, it will serve as a reference for new developers to understand the available UI components and their variants quickly.
The page will display a list of components, each labeled with a corresponding header, and render the component in several variants. 

## Definition of Done
1. A dedicated route (e.g., /components-test) has been created in the app.
2. The page includes headers labeling each component, with the component displayed below each header.
3. Each component is rendered in multiple variants (different sizes, colors, states, etc., as applicable).

mackcooper1408:
I think we can actually unblock this and build the initial page into the router. Then, as part of the DoD ("Definition of Done") for each component we build we can say "add this component to the test page for people to see/review"

PR #51, opened 2024-10-02
Component test page

mantuok:
# Description

Created an initial page for the components test library and added a mock button as a reference example.

#39 

## Type of changes
- [ ] Bugfix
- [ ] Chore
- [x] New Feature

## Testing
- [ ] I added automated tests

### DIFF
--- a/app/components-test-lib/page.tsx
+++ b/app/components-test-lib/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { Heading, useToast } from "@chakra-ui/react";
+import { Box, Heading, VStack, useToast, Text, HStack } from "@chakra-ui/react";
 import MockButton from "../components/mock-button";
 
 const ComponentsTestLib = () => {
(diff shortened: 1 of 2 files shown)


app/components-test-lib/page.tsx
```diff
@@ -0,0 +1,67 @@
+"use client";
+
+import { Box, Heading, VStack, useToast, Text, HStack } from "@chakra-ui/react";
+import MockButton from "../components/mock-button";
+
+const ComponentsTestLib = () => {
+  const toast = useToast();
+
+  if (process.env.NODE_ENV !== "development") {
```
mackcooper1408, line 9: i actually think we can get rid of this because we kind of _do_ want this in "production" (or at least in the sandbox environment we setup for testing), that way non-eng stakeholders can look at the page and see what we've got going on

app/components-test-lib/page.tsx
```diff
@@ -0,0 +1,67 @@
...
+      title: "Action Successful.",
+      description: message,
+      status: "success",
+      duration: 5000,
+      isClosable: true,
+    });
+  };
+
+  return (
+    <>
```
mackcooper1408, line 24: nit: don't need the fragment since the box wraps everything

mackcooper1408 (approved):
love it! 🎨

Issue #27, opened 2024-09-14
Configure App Environments


mackcooper1408:
There's not much to add here until these issues (#22 and #23 are done)

## Context
In order to deploy our application and work with containerization tools like Docker, we'll need some extra configuration to use certain Environment variables. The easiest option for making this work is via a secret `.env` file that will not be checked into git version history.

## Definition of Done
1. environment configuration is complete for front and back end with the app now pulling key variables like "Environment", "Database URL", etc. from an "ENV" file (`.env`)
2. env configuration has appropriate defaults for working with the app in development
3. `.gitignore` is updated to not check the env file into git history
4. README is updated with details on how this config is setup and what variables may be useful (if any at this point) to add to your local `.env`

## Engineering Details
- see [this doc](https://fastapi.tiangolo.com/advanced/settings/) for details on completing the setup in FastAPI.
  - Also, take a look at how the configuration is done in this [template app](https://github.com/fastapi/full-stack-fastapi-template)
- it [appears](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables) NEXT has native support for .env files... so yay!

agennadi:
Done

agennadi:
Done

PR #66, opened 2024-10-22
Configure app envs

agennadi:
# Description

Environment variables are stored centrally in the `.env.local` file. 
The `compose.yaml` is updated to use variables from `.env.local`.
The `Settings` class is created for Pydantic to read the environment variables.
`README` is updated.

#27 

## Type of changes
- [ ] Bugfix
- [ ] Chore
- [x] New Feature

## Testing
- [ ] I added automated tests

## How to test

Run docker-compose up -d
Access the application at http://localhost:3000/
Access the API at http://localhost:8000/docs
Run pytest backend/database/tests/test_database.py to test the database
Run docker-compose down


mackcooper1408 (approved):
Looks good! Check GH, seems like it's mad about you not having enough empty lines at the end of files haha. Also, wasn't able to run pytest... do I still need to set up a virtual env and install packages...? or is that supposed to be handled by docker? Probably just something on my end haha just checking

Issue #71, opened 2024-11-02
Set up a SQLAlchemy model and an API for Addresses 


agennadi:
Context
-------
We need to establish a new SQLAlchemy model for an Addresses entity based on the SFData dataset, specifically the [SF Addresses dataset](https://data.sfgov.org/Geographic-Locations-and-Boundaries/Addresses-with-Units-Enterprise-Addressing-System/ramy-di5m/about_data). This model will store details about each address in San Francisco, including geospatial information to support location-based queries. For efficient geospatial handling, a Point type field should be included in the model. Additionally, the Docker environment should support the PostGIS extension to handle spatial data.

Definition of Done
------------------

-   PostGIS extension is supported by Docker. 
-   SQLAlchemy model for the Addresses entity is created with fields that align with the SFData dataset structure.
-   Pydentic schema for Addresses is created to validated API outputs.
-   Get API for Addresses is created.
-  Tests for the API are created and run successfully.  

Engineering Details
-------------------

-  SF Addresses dataset: https://data.sfgov.org/Geographic-Locations-and-Boundaries/Addresses-with-Units-Enterprise-Addressing-System/ramy-di5m/about_data
- The official PostGIS Docker image currently lacks native support for ARM64 architecture. To address this, a custom PostGIS image should be built locally, configured for ARM architecture to support M1/M2 chipsets. This local build will ensure compatibility and allow development to proceed without errors related to spatial operations.

agennadi:
Done

PR #76, opened 2024-11-06
Add a model, a schema, and an API for addresses

agennadi:
# Description

Created a model, a Pydantic schema, an API for Addresses.
Changed Postgres Image to a locally built PostGIS image.
Many of the file changes are caused by linting. 

#71

## Type of changes
- [ ] Bugfix
- [ ] Chore
- [X] New Feature

## Testing
- [X] I added tests

## How to test

Run docker-compose up -d
Access the application at http://localhost:3000/
Access the API at http://localhost:8000/docs
Run `pytest backend/api/tests/test_addresses.py` in the terminal to test the GET API for Addresses.
Run docker-compose down.


backend/api/models/addresses.py
```diff
@@ -0,0 +1,54 @@
...
+from .base import Base
+
+"""Data from https://data.sfgov.org/Geographic-Locations-and-Boundaries/Addresses-with-Units-Enterprise-Addressing-System/ramy-di5m/about_data"""
+
+
+MAPPED_COLUMN_STRING_LENGTH = 255
+
+
+class Address(Base):
+    __tablename__ = "address"
```
mackcooper1408, line 17: do we want this to be plural and match the file name (so `addresses`)

backend/api/models/addresses.py
```diff
@@ -0,0 +1,54 @@
...
+    parcel_number: Mapped[str] = mapped_column(String)
+    block: Mapped[str] = mapped_column(String)
+    lot: Mapped[str] = mapped_column(String)
+    cnn: Mapped[int] = mapped_column(Integer)
+    longitude: Mapped[float] = mapped_column(Float, nullable=False)
+    latitude: Mapped[float] = mapped_column(Float, nullable=False)
+    zip_code: Mapped[int] = mapped_column(Integer, nullable=False)
+    """point: Mapped[Geometry] = mapped_column(
+        Geometry("POINT", srid=4326, spatial_index=True), nullable=False
+    )"""
```
mackcooper1408, line 36: is this commented out for good? or are we expecting to add this back in?

backend/api/models/landslide_zones.py
```diff
@@ -1,33 +1,40 @@
...
 
-class LandslideZones(DeclarativeBase):
+class LandslideZone(Base):
     """
     All data of the Landslide Zones table from SFData.
     Contains multipolygon geometries defining landslide zones.
     """
 
-    __tablename__ = "landslide_zones"
+    __tablename__ = "landslide_zone"
```
mackcooper1408, line 18: what's the main reason for making this singular?

backend/api/models/liquefaction_zones.py
```diff
@@ -1,26 +1,31 @@
...
 from sqlalchemy.orm import DeclarativeBase
 from sqlalchemy.orm import Mapped
 from sqlalchemy.orm import mapped_column
 from geoalchemy2 import Geometry
 from datetime import datetime, DateTime
+from .base import Base
 
 
-class LiquefactionZones(DeclarativeBase):
+class LiquefactionZons(Base):
```
mackcooper1408, line 12:
```suggestion
class LiquefactionZones(Base):
```

backend/api/models/liquefaction_zones.py
```diff
@@ -1,26 +1,31 @@
...
     """
     All data of the Liquefaction Zones table from SFData.
-    Contains multipolygon geometries defining soil liquefaction zones as High (H) or 
+    Contains multipolygon geometries defining soil liquefaction zones as High (H) or
     Very High (VH) susceptibility.
     """
-    __tablename__ = "liquefaction_zones"
+
+    __tablename__ = "liquefaction_zone"
```
mackcooper1408, line 19: same q about keeping this plural

mackcooper1408 (approved):
overall looks great! a couple of little comments and questions. Biggest theme is basically:

1. are we moving to a pattern with singular named database tables? I personally like tables in the DB with plural names... we could keep the class names singular (which I also prefer) but the `__tablename__` plural. but not a huge deal, let's just make sure we're consistent. Also, if everything is singular, let's also change the file names.
2. It looks like we're pulling all the fields from the data source... I'm assuming this is temporary until we hone in on what we actually want to use...? seems like we probably don't need all those fields...
3. I couldn't actually run this locally... the postgis db Error'd out... (see image)
Screenshot 2024-11-06 at 7 36 48 PM

Issue #128, opened 2024-12-23
Create APIs for address risk


elucherini:
## Context
For the MVP, we need to support this user flow:
1. User searches an address through search box.
2. User gets back risk(s) for that address.

This is the **backend** issue for this user flow. We should also open a corresponding frontend issue, if it doesn't already exist.

## Definition of Done
Three APIs with related tests:

- [x] API that, given an address, tells whether it is a soft story
- [x] API that, given an address, tells whether it is in a ~~liquefaction~~ seismic zone
- [x] API that, given an address, tells whether it is in a zone at risk of tsunami

Translating the address obtained through MapBox autocomplete into a geometrical Point could be handled by the frontend (so we keep all MapBox-related stuff in the frontend), so it is currently out of scope for this issue.

## Engineering Details
All these APIs are GET requests that take latitude and longitude as input. The exact output should be further discussed with the frontend team. For now, all APIs return a simple boolean.

PR #130, opened 2024-12-23
Create APIs for address risk

elucherini:
# Description

Implementing issue #128.

Three APIs with related tests:

- [x] API that, given an address, tells whether it is a soft story
- [x] API that, given an address, tells whether it is in a liquefaction zone
- [x] API that, given an address, tells whether it is in a zone at risk of tsunami

Translating the address obtained through MapBox autocomplete into a geometrical Point could be handled by the frontend (so we keep all MapBox-related stuff in the frontend), so it is currently out of scope.

---

**Also added pre_commit to requirements-dev.txt.**

## Type of changes
- [ ] Bugfix
- [ ] Chore
- [x] New Feature

## Testing
- [x] I added automated tests

## How to test

- Run app

### Soft-story
- Go to http://localhost:8000/soft-stories/is-soft-story?lat=-122.446575165&lon=37.766034349 and see that the response is `true`. This is a coordinate in our soft-story table.
- Go to http://localhost:8000/soft-stories/is-soft-story?lat=0&lon=0 and see that the response is `false`. This is Null Island, which is not in our table.

### Liquefaction
- Go to http://localhost:8000/liquefaction-zones/is-in-liquefaction-zone?lat=37.779759&lon=-122.407436 and see that the response is `true`. This is a point at 6th and Howard in a liquefaction zone.
- Go to http://localhost:8000/liquefaction-zones/is-in-liquefaction-zone?lat=0&lon=0 and see that the response is `false`. This is Null Island. Maybe it is in a liquefaction zone, but it's certainly out of our jurisdiction.

### Tsunami
- Go to http://localhost:8000/tsunami/is-in-tsunami-zone?lat=37.759039&lon=-122.509515 and see that the response is `true`. This is a point in the Outer Sunset in our table.
- Go to http://localhost:8000/tsunami/is-in-tsunami-zone?lat=0&lon=0 and see that the response is `false`. This is Null Island. Maybe it is at Tsunami risk but it's certainly out of our jurisdiction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ node_modules
 .env.development.local
 .env
 .env.local
+
+# IDEs
+.vscode/
--- a/backend/api/routers/liquefaction_api.py
+++ b/backend/api/routers/liquefaction_api.py
@@ -45,7 +45,9 @@ async def get_liquefaction_zones(db: Session = Depends(get_db)):
 
 
 @router.get("/is-in-liquefaction-zone", response_model=bool)
-async def is_in_liquefaction_zone(lat: float, lon: float, db: Session = Depends(get_db)):
+async def is_in_liquefaction_zone(
+    lat: float, lon: float, db: Session = Depends(get_db)
+):
     """
     Check if a point is in a liquefaction zone.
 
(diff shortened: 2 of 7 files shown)


elucherini:
@agennadi docker build failing due to env-file missing in the artifacts?

elucherini:
Closing as the CI pipeline doesn't work from forks.

Issue #138, opened 2024-12-30
MapBox coordinate checks


leela-solomon:
- Test that MapBox coordinates match the in-table coordinates for soft-story street addresses.
- Test that MapBox polygon edges are classifying correctly. (How to verify that it's correct or not? We may have to ensure that we are using the correct coordinate display settings that the polygon was originally encoded for.)

elucherini:
Some more info in issue #129, created last week.

Issue #124, opened 2024-12-19
Decide how to handle null coordinate point entries in the Soft-Story properties dataset


agennadi:
## Context
The Soft-Story properties dataset currently contains 4941 rows, with approximately 10 rows having null values in the `point` field. Valid point values are essential for accurately displaying soft-story properties on the map. These missing points need to be resolved during the transformation process of the dataset.

One potential approach is to load [Addresses dataset](https://data.sfgov.org/Geographic-Locations-and-Boundaries/Addresses-with-Units-Enterprise-Addressing-System/ramy-di5m/about_data) to look up the missing points based on property addresses. However, loading 400,000 addresses to resolve 10 missing points may be excessive.

Alternatively, the missing points could be added manually. While this might work for now, it is not a scalable solution for future cases.

Research is needed to identify the optimal method for handling null values in the dataset.

## Definition of Done
- The optimal solution for handling null values is identified.
- The solution is shared with the backend team.

## Technical Details
Check out the [Soft-Story properties dataset](https://data.sfgov.org/Housing-and-Buildings/Soft-Story-Properties/beah-shgi/about_data).
Relevant ETL code can be found in `backend/etl/soft_story_properties_data_handler.py`.

leela-solomon:
1. Add source column to track the nulls which we filled in.
2. Fill in the null coordinates manually or with easy automation.

elucherini:
We can write manual UPDATE statements to fill in the nulls, making sure we also update the source column. Alternatively, we could keep a separate table and left-join it to the original, but then we end up with two tables to maintain. I think option 1 is good enough for 10 missing points.

The data source can be MapBox. The simplest solution is to use this: https://docs.mapbox.com/playground/geocoding/ , or have ChatGPT write a simple script using our API key.

elucherini:
All 10 addresses with missing coordinates in the SFData soft story dataset are resolvable by MapBox. Since we already added the MapBox geojson as a data source in #142, we won't need to write manual UPDATE statements to fill in the nulls. I can simply modify the query in `get_soft_stories` to return the rows with either field `point` or `mapbox_point` is populated.

The issue now is the FE team needs to know that, when `point` is NULL, they need to get coordinates from field `mapbox_point` (cc @nickvisut). This might not be the ideal solution to communicate with the front-end.

Alternatives are to change my code in #142 to instead add a `point_source` field, as originally discussed. Then:

[...]

PR #142, opened 2025-01-06
Add soft story Mapbox integration

elucherini:
# Description

Mapbox's set of coordinates is inherently different from the one provided by SF Data. For soft story lookup, we need a 1:1 match, so we add a Mapbox soft story dataset and manager class.

Addresses both #138 and #124:

1. It solves the issue in #138
2. It overwrites SFData coordinates with Mapbox coordinates, whenever possible, so #124 is also addressed.

## Implementation notes

In the soft story ETL, before populating the database:

- [x]  We fetch and parse the SFData soft story dataset and add a MapboxGeojsonManager object to the ETL. Internally, when the MapboxGeojsonManager is initialized, it reads the raw Mapbox geojson and creates a dictionary `{sfdata_address: (mabox_lon, mapbox_lat)}` (see last point to see how sfdata_address is added to the Mapbox geojson).
- [x] At each SFData address, we try to get a set of Mapbox coordinates from the MapboxGeojsonManager; if found, `point` is set to the Mapbox coordinates; if not, `point` is set to the SFData coordinates, or None if those aren't available, either.
- [x] After processing all SFData soft story data, we collect all addresses where Mapbox coordinates weren't defined and geocode them in batch by calling the Mapbox API. Then, we fill all `point` fields for whatever coordinates we found. Note that Mapbox won't resolve all addresses in the dataset, so some of these will still be None. This problem needs to be documented.
- [x] The new data fetched from Mapbox, if any, is also appended to the geojson. For each geojson feature, we add a field "sfdata_address" to key "properties" to easily join the Mapbox data with the SFData soft story dataset. In other words, we expect all features stored in the geojson to have this structure:

```
 {
            "type": "FeatureCollection",
            "features": [
                {
                    "properties": {
                        "sfdata_address": "Address 1",
                        "coordinates": [
                            "longitude": -122.4,
                            "latitude": 37.8
                         ],
                        # other fields...
                    }
                },
              # other features...
            ],
```

## Type of changes
- [x] Bugfix
- [ ] Chore
- [x] New Feature

## Testing
- [x] I added automated tests

## How to test

Run tests.


backend/etl/soft_story_properties_data_handler.py
```diff
@@ -65,12 +61,17 @@ def parse_data(self, data: dict) -> list[dict]:
             geometry = feature.get("geometry", {})
             if geometry:
                 geom_longitude, geom_latitude = geometry["coordinates"]
-            addresses.append(properties.get("address"))
 
             # Search for the address in the mapbox geojson
-            mapbox_coordinates = self.mapbox_geojson_manager.get_mapbox_coordinates(
+            if not self.mapbox_geojson_manager.is_address_in_geojson(
```
elucherini, line 66: If not found in the geojson, store to fill in later -- this is reserved for new addresses in the SFData dataset.

backend/etl/soft_story_properties_data_handler.py
```diff
@@ -65,12 +61,17 @@ def parse_data(self, data: dict) -> list[dict]:
...
                 properties.get("address")
-            )
+            ):
+                addresses.append(properties.get("address"))
+                mapbox_coordinates = None
+            else:
+                mapbox_coordinates = self.mapbox_geojson_manager.get_mapbox_coordinates(
```
elucherini, line 72: If the address is in the dataset, retrieve its value. Note that this function might still return None, if the address can't be resolved by Mapbox. In our 1000 soft story sample, this happens for 56 addresses.

backend/api/routers/soft_story_api.py
```diff
@@ -45,17 +47,19 @@ async def get_soft_stories(db: Session = Depends(get_db)):
...
+        db (Session): The database session dependency
 
     Returns:
-        bool: True if the point is a soft story property, False otherwise.
+        bool: True if the point is a soft story property, False
+        otherwise
     """
     query = db.query(SoftStoryProperty).filter(
-        SoftStoryProperty.point == geo_func.ST_GeomFromText(f"POINT({lon} {lat})", 4326)
+        SoftStoryProperty.mapbox_point
```
AdamFinkle, line 62: Eli and I decided to let the function to return False for any address without a MapBox point because trying to match MapBox coordinates retrieved from an entered address to any SFData point would fail anyway because the precision and values differ.  In the geojson are 56 addresses which MapBox could not resolve.  Our best guess, based on this number of 1,000 soft story addresses queried from SFData, is that about 200 addresses of the about 4,000 soft story addresses.

We didn't change get_soft_stories because SFData lacks coordinates for only 10 addresses while MapBox is missing more.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: 23.3.0
+    rev: 24.10.0
```
elucherini, line 3: Fixing precommit

agennadi (approved):
Looks great

PR #161, opened 2025-01-20
Load data into Vercel using GitHub Actions

### LINKED ISSUE #150: Project account: Load tables into Vercel and create ETL pipeline

### LINKED ISSUE #154: Optimize Handling of Large Multipolygons
## Context
The current dataset includes extremely large multipolygons (especially in the Liquefaction dataset) that cause performance bottlenecks in API calls. Users experience timeouts or delays (>10 seconds) when querying these geometries.

While simplifying geometries is a primary approach for now, other optimization strategies could also help reduce complexity and improve performance without compromising data integrity. These should be evaluated as part of the solution.

## Definition of Done
- Large multipolygons are optimized using one or more of the following approaches:
  - Simplification using PostGIS `ST_Simplify` with the tolerance level of `0.00001`.
  - Splitting large multipolygons into smaller, manageable chunks using `ST_Subdivide` or custom logic.
  - Compression techniques, such as converting geometries to a lighter format like GeoJSON with reduced precision.
  - Filtering out redundant or unnecessary vertices using `ST_RemoveRepeatedPoints`.
  - Other possible approaches? 
- PostGIS spatial indexes (e.g., `GIST`) are added to optimize queries involving geometries.
- The API is updated to handle optimized geometries without breaking existing functionality.
- Tests confirm that database query times for spatial operations are significantly improved.

## Engineering Details
This is how `ST_Simplify` simplifies polygons: 
- Original polygon: Many points are very close together (e.g., 0.000005 degrees apart).
- Simplified polygon: Points that are closer than 0.00001 degrees are collapsed into a single point, slightly smoothing the shape but keeping the overall structure intact.

See https://postgis.net/docs/ST_Simplify.html for details.

Note that length and area of polygons also change when you simplify them.

agennadi:
# Description

- A GItHub Actions workflow that populates Tsunami, Soft Story, and Liquefaction tables is created.
- The workflow runs on a schedule and can also be triggered manually.
- The Vercel's database is populated with the Soft-Story Properties, Tsunami, and Liquefaction datasets
- A UNIQUE constraint in the Soft Story table was added and the ETL scripts were modified to avoid duplicates in the database.
- Multipolygons in the Liquefaction table are simplified using PostGIS `ST_Simplify` with the tolerance level of 0.0001

Closes #150, addresses but doesn't close #154 

## Type of changes
- [ ] Bugfix
- [ ] Chore
- [X] New Feature

## Testing
- [ ] I added automated tests

## How to test
Manually trigger the `ETL to Neon` workflow.


.github/workflows/etl_to_neon.yml
agennadi, line 1: While in development, the workflow is only triggered once a week - to save on data transfer costs. The free limit of data transfer is 5 GB/month.

backend/etl/soft_story_properties_data_handler.py
agennadi, line 1: Got a mypy error: `Incompatible types in assignment (expression has type "dict[str, tuple[float, float] | None]", variable has type "dict[str, tuple[float, float]]")`. Wasn't sure if I can safely change the type, so temporarily silenced the warning.

backend/etl/liquefaction_data_handler.py
agennadi, line 1: The duplicate check relied on the auto increment PK, which prevented duplicates from being detected. Replaced the autoincrement key with a composite PK

backend/api/models/soft_story_properties.py
agennadi, line 1: Added a `UNIQUE` constraint to the `Property_address` field, so that it can be used to prevent duplicates in the `Soft_story_properties` table.

.github/workflows/etl_to_neon.yml
elucherini: Is this a neon or Github Actions limit? Just curious.

backend/database/session.py
```diff
@@ -3,7 +3,7 @@
 from backend.api.config import settings
 
 # Set up the database engine using settings
-engine = create_engine(settings.database_url_sqlalchemy, echo=True)
+engine = create_engine(settings.neon_url, echo=True)
```
elucherini, line 6: Are there consequences to changing this script to connect to the Neon DB as opposed to the local DB? As in, if I want to test my local DB, do I need to change this line back to connect to `database_url_sqlalchemy`?

elucherini (commented):
Only one concern, everything else looks good. I will take a look at the mapbox geojson manager type error.

.github/workflows/etl_to_neon.yml
agennadi: This is a neon limit

backend/database/session.py
agennadi: Yes, you'd have to change this line back to test locally. If this is a problem, I can add an if-clause that would switch the url based on the environment

elucherini: Ok, I think it’s fine for now. I would probably prefer putting more data in the test DB.

elucherini (approved):

Issue #159, opened 2025-01-18
Clean soft story addresses before querying Mapbox during ETL


elucherini:
## Context
Mapbox can't resolve 61 addresses out of our soft story dataset. For these addresses, we are using the default SF Data address, so we can populate the map whenever needed. However, without a Mapbox address, we lose the lookup functionality for these soft stories.

Many (if not all) of these unresolvable addresses are in this format: `X 09th Ave`, or similar. We think removing the leading zeros can help get more addresses from Mapbox.

## Definition of Done

- [x] Quantify SF Data addresses that fall in the leading-zero category -- that is, the format is, for example `X 09th Ave`.
- [x] Quantify leading-zero addresses Mapbox cannot translate; presumably, Mapbox cannot translate any of these.
- [x] In the ETL, write code to lightly clean these SF Data addresses before querying Mapbox by removing the leading zeros.
- [x] Rerun the ETL and quantify addresses left that Mapbox cannot translate; presumably, this will be < 61 addresses Mapbox can't currently translate.

## Engineering Details
Hopefully "Definition of Done" gives a pretty exhaustive overview of what is needed here.

elucherini:
These changes resulted in a total of two unresolvable addresses instead of 61 out of 1000 soft stories:

2 LEONA TR, SAN FRANCISCO CA
10 FLORENTINE AV, SAN FRANCISCO CA

We confirmed that these (or reasonable variations) were not available in Mapbox, by using: https://docs.mapbox.com/playground/geocoding/.

However, we noticed that two other addresses are reported by Mapbox to be in other cities:
1. `545 ARGUELLO ST, SAN FRANCISCO CA` resolves to 545 Arguello Street, Redwood City. The SF address just isn't available in Mapbox, so the desired result would be None.
2. `450 CASTRO ST, SAN FRANCISCO CA` resolves to 450 Castro Street, San Leandro. The SF address is available in Mapbox, but appears to be the third match, for some reason.

Mapbox allows adding proximity to a set of coordinates in the [geocoding input](https://docs.mapbox.com/api/search/geocoding/#forward-geocoding-with-search-text-input). Adding the SF data coordinates (whenever available) didn't result in an improvement.

The most reasonable idea (still to be tested) is adding a boundary box around SF and adding that to the input with the `bbox` parameter.

Of course, these two addresses can also be dropped. However, this should be done only once we have an idea of how many rows we are going drop in the total soft story properties dataset of ~5k entries.

[...]

PR #169, opened 2025-01-26
Add bounding box to mapbox query

elucherini:
# Description

Adding a bounding box for the city of San Francisco. See gist in the wiki for how the bounding box is calculated (TL;DR: by taking the min and max longitudes, and min and max latitudes in the SFData address dataset) and verified.

In practice, the bounding box prevents mapbox from matching addresses outside of San Francisco (#159). This issue affected only two addresses out of 1000, but it was unclear how large of a problem it would have been on the full 4.9k soft story property dataset.

As we expand, we might need to add more bounding boxes for the new geographies.

## Type of changes
- [x] Bugfix
- [ ] Chore
- [ ] New Feature

## Testing
- [ ] I added automated tests

## How to test

Run the ETL, then:

```
docker exec -it my_postgis_db psql -U postgres -d qsdatabase

qsdatabase=# SELECT address, point, point_source FROM soft_story_properties WHERE point_source != 'mapbox';
```

See that there's only one row (this address cannot be found in Mapbox)

```
qsdatabase=# SELECT address, point, point_source FROM soft_story_properties WHERE point_source != 'mapbox';
           address            |                       point                        | point_source 
------------------------------+----------------------------------------------------+--------------
 2 LEONA TR, SAN FRANCISCO CA | 0101000020E6100000B285FFF87B9C5EC060C36E003FE44240 | sfdata
(1 row)
```


agennadi (approved):
Lgtm

Issue #168, opened 2025-01-23
'npm next-dev` development server not working


nickvisut:
Running `npm run next-dev` or `npm run dev` errors out (https://localhost:3000 doesn't properly show the app). Last known good build is `2cf779d32ebf4601bdbf7fe7f6ed041a5b9243d1`.

Add'l details from @nladio:
> I checked out the commit before 5f460... - 3a6d057afabe55b4862a8a7fb915ff11ef942076 , and tried using both the setup methods without success.
> When I visit localhost:3000, the frontend service tries to load the page for a couple of minutes, and then crashes. I see this same error in the logs
> [0]  ○ Compiling / ...
> [0]  ✓ Compiled / in 11.5s (3012 modules)
> [0] npm run next-dev exited with code 0

Full convo here: https://sfbrigade.slack.com/archives/C07DB8HBV9P/p1737431256673869 (TODO: paste entire thing in).

nickvisut:
@nladio generated a potential workaround here: https://github.com/sfbrigade/datasci-earthquake/pull/163 that could give us clues on how to ultimately fix it.

nickvisut:
Thanks @nladio and @agennadi for reporting the issue.

PR #172, opened 2025-01-28
Bug Fix: npm run next-dev Failing to Start

mantuok:
# Description

Fixed an issue where npm run next-dev was not working. The problem was caused by upgrading Next.js to version 15. Downgraded Next.js to version 13, and the development server now runs successfully.

#168 

## Type of changes
- [x] Bugfix
- [ ] Chore
- [ ] New Feature

## How to test
Run `npm install,` then  `npm run next-dev`


agennadi (approved):

nickvisut:
good catch! will investigate what's causing newer Next to blow up.

Issue #182, opened 2025-01-31
Local environment should not point to prod (Neon) DB


elucherini:
## Context
The environment variables point to the prod DB (Neon DB) by default. However, ETLs started locally (or anything run locally, for that matter) should write locally by default. Only ETLs ran through CircleCI should write to prod. Otherwise, the risk of overwriting the prod DB is too high, especially as we inch closer to deployment.

## Definition of Done
Running the ETL locally won't affect the prod environment.

## Engineering Details
...

PR #187, opened 2025-02-01
Default to local DB when running ETL locally

elucherini:
# Description

This PR allows us to default to writing to the local DB when running the ETL process locally. 

Addresses #182.

It requires a GitHub secret change: ENVIRONMENT should be set to `ci` or `prod`. When generating the .env file, the ENVIRONMENT variable will be set to "local". It is currently in plain text.

Security-wise, this approach should be okay because:

1. We are not exposing a token, key, or host name, just an environment variable that is meaningless without the actual tokens and host names.
2. By manually changing the variable to "ci", an unauthorized user would only be able to create a session with the dummy NEON_URL variable in the .env.example, which still wouldn't allow them to write to production.

On the other hand, it is an easy change for an authorized user who needs to write to production bypassing GitHub action.

## Type of changes
- [x] Bugfix
- [ ] Chore
- [ ] New Feature

## Testing
- [ ] I added automated tests

## How to test

Run ETL locally, see that it writes to the local DB and not the Neon DB.

## Clean commits

### DIFF
--- a/.github/workflows/env_vars.yml
+++ b/.github/workflows/env_vars.yml
@@ -24,7 +24,7 @@ jobs:
         envkey_DATABASE_URL_SQLALCHEMY: ${{ secrets.DATABASE_URL_SQLALCHEMY }}
         envkey_DATABASE_URL_SQLALCHEMY_TEST: ${{ secrets.DATABASE_URL_SQLALCHEMY_TEST }}
         envkey_LOCALHOST_DATABASE_URL_SQLALCHEMY: ${{ secrets.LOCALHOST_DATABASE_URL_SQLALCHEMY }}
-        envkey_ENVIRONMENT: ${{ secrets.ENVIRONMENT }}
+        envkey_ENVIRONMENT: local
         envkey_SECRET_KEY: ${{ secrets.SECRET_KEY }}
 
         envkey_NEXT_PUBLIC_API_URL: ${{ secrets.NEXT_PUBLIC_API_URL }}
--- a/backend/database/session.py
+++ b/backend/database/session.py
@@ -2,8 +2,19 @@ from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 from backend.api.config import settings
 
+
+def _get_database_url() -> str:
+    match settings.environment:
+        case "local":
+            return settings.database_url_sqlalchemy
+        case "ci" | "prod":
+            return settings.neon_url
+        case _:
+            raise ValueError(f"Unknown environment: {settings.environment}")
+
+
 # Set up the database engine using settings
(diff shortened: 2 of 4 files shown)


backend/etl/soft_story_properties_data_handler.py
```diff
@@ -6,7 +6,6 @@
 from dotenv import load_dotenv
 import os
 from pathlib import Path
-from typing import Dict, Tuple
```
elucherini, line 9: Just an extra thing I noticed, has nothing to do with PR itself.

backend/database/session.py
```diff
@@ -2,8 +2,19 @@
 from sqlalchemy.orm import sessionmaker
 from backend.api.config import settings
 
+
+def _get_database_url() -> str:
+    match settings.environment:
+        case "local":
+            return settings.database_url_sqlalchemy
+        case "ci" | "prod":
```
elucherini, line 10: Requires GitHub secret to be changed to `ci` for the "ETL to Neon" script to work.

.github/workflows/env_vars.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         envkey_DATABASE_URL_SQLALCHEMY: ${{ secrets.DATABASE_URL_SQLALCHEMY }}
         envkey_DATABASE_URL_SQLALCHEMY_TEST: ${{ secrets.DATABASE_URL_SQLALCHEMY_TEST }}
         envkey_LOCALHOST_DATABASE_URL_SQLALCHEMY: ${{ secrets.LOCALHOST_DATABASE_URL_SQLALCHEMY }}
-        envkey_ENVIRONMENT: ${{ secrets.ENVIRONMENT }}
+        envkey_ENVIRONMENT: local
```
elucherini, line 27: Since I set "environment" in the settings to default to "local", another approach is to remove it altogether from .env generation if we really don't want this to be in plain text. However, since the pattern matching is the code is in plain text I don't see much of a downside to this.

agennadi (approved):
Looks good! I changed the `environment` secret to `ci`. Could you please keep the imports absolute? The `etl_to_neon` pipeline breaks because `soft_story_properties_data_handler.py` imports from `etl.mapbox_geojson_manager` instead of `backend.etl.mapbox_geojson_manager`

elucherini:
Thanks, Anna!

> Could you please keep the imports absolute? The `etl_to_neon` pipeline breaks because `soft_story_properties_data_handler.py` imports from `etl.mapbox_geojson_manager` instead of `backend.etl.mapbox_geojson_manager`

Sure, I'll change that. This is solvable by adding the backend directory to the environment variable PYTHONPATH, btw. Although in this case I don't think it matters much to use relative paths (we are not creating a package or anything).

Issue #189, opened 2025-02-02
[ETL] Remove the unused data handlers.


nladio:
## Context

There are currently 6 ETL data handler flows:

- Liquefaction
- Soft story
- Tsunami
- Landslide
- Address
- Seismic

We use the first three, and intend to use the fourth after the MVP. We can cleanup the code for the last two as they're not intended to be used.

## Definition of Done

- Remove the address and seismic data handler jobs.
- Update the DB schema to remove their tables.
- Remove the models and routers.
- Ensure the entire stack (DB + BE + FE) works as expected after the code removal.

## Engineering Details

All the data handler files are located under `backend/etl`. 
You can review the readme to get the latest instructions to run the stack locally.

PR #204, opened 2025-02-08
Delete addresses and seismic

AdamFinkle:
# Description

Deleted the address and seismic tables from the database alongside all ancillary files.

#189

## Type of changes
- [ ] Bugfix
- [x] Chore
- [ ] New Feature

## Testing
- [ ] I added automated tests

## Clean commits


requirements-dev.txt
agennadi, line 1: Do we actually use these requirements anywhere? Is there a reason why `requirements.txt` isn't enough?

agennadi:
Looks good!

The tests fail because there's an error in the `init.sql`. You didn't remove the `seismic_hazard_risk` field when inserting values into the `combined_risk` table. 

In fact, we don't need the `combined_risk` table as well. You could remove it in this PR or in a separate one. If you prefer to open a separate issue, you can just comment out the insert statement and the tests will pass

AdamFinkle:
I have removed the combined risk table from `init.sql`

agennadi (changes requested):
Since you removed the `combined_risk` table from `init.sql`, the tests related to the `combined_risk` table are failing. The tests pass if you keep the table but remove the insert statements. Please make sure to test locally before pushing the changes. If tests fail, the first step is to check the container logs.

Issue #238, opened 2025-02-20
[CI] Ensure all the containers are healthy after they startup


nladio:
**Describe the bug**
Currently, our CI [workflow](https://github.com/sfbrigade/datasci-earthquake/blob/724e36f6237b05b1f5ee40134188016c3922722f/.github/workflows/ci.yml#L95) runs the unit tests to confirm if a docker container is healthy.
But we have low unit test coverage, and sometime our test pass but the ETL data_handler scripts may fail. 

**To Reproduce**
This issue was initially noticed by @agennadi, for a bug which was fixed in https://github.com/sfbrigade/datasci-earthquake/pull/201. If you try to revert that commit and create a new draft PR you might be able to repro the issue.

**Expected behavior**
It takes around ~30s to run the startup scripts for the backend container. Not sure if the other containers have startup scripts longer than this. We should ensure the docker containers are still healthy after a couple of minutes. 

Add a docker health check run on all the containers a couple minutes after the containers startup (this could be as low as 45s, depending on how quickly our services startup). https://lumigo.io/container-monitoring/docker-health-check-a-practical-guide/

gamesover:
@nladio 

I believe that allowing 30 seconds for container startup is quite reasonable; however, I have noticed that many docker-compose configurations use a 60-second start period for services. I'll adjust the backend startup wait time to 45 seconds. This should provide enough time for the service to stabilize while avoiding unnecessary delays in our CI workflow.

PR #252, opened 2025-02-27
174 - add health check for docker containers

gamesover:
# Description

- Add healthcheck for frontend and backend containers
- CI needs to wait for all containers healthy before testing

#238 

## Type of changes
- [ ] Bugfix
- [x] Chore
- [ ] New Feature

## Testing
- [ ] I added automated tests

## How to test

As long as passing CI

## Clean commits

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Get Run ID of Most Recent Successful Run
         id: get_run_id
@@ -72,10 +72,10 @@ jobs:
           echo "Decryption complete"
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
 
       - name: Cache Docker layers
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: /tmp/.buildx-cache
           key: ${{ runner.os }}-buildx-${{ github.sha }}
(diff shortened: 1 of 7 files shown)


gamesover:
Actually, I was able to reproduce the CI issue locally. Previously, I believed it was due to a problem with my local environment configuration. Let me fix the issue properly.

Dockerfile
agennadi, line 1: Why do we need to explicitly create the `.next` folder if it is automatically created by `next dev`?

.github/workflows/ci.yml
agennadi, line 1: Are 90s sufficient? Especially if the start_period of the frontend container is 3m?

agennadi (approved):
Looks good!

Dockerfile
gamesover: I asked claude 3.7 the same question :)

below it is AI answer

# Creating the .next Directory in Dockerfile

You're absolutely right to question this. Creating the .next directory explicitly in the Dockerfile is not strictly necessary since Next.js will create it automatically when you run `next dev` or during the build process.

## Why it's done in this Dockerfile:

1. **Permission management**: The main reason is to set permissions on this directory before Next.js tries to use it. By creating it beforehand and setting the permissions to `777` (full read/write/execute for everyone), you're ensuring that the non-privileged user (`appuser`) will have full access to write to this directory.

2. **Avoiding permission errors**: Without this step, Next.js might create the .next directory with default permissions that could cause the EACCES errors you were seeing earlier, especially when running as a non-root user in a Docker container.

This is essentially a preemptive fix for the common permission issues that arise when running Next.js in containerized environments with non-root users. The error you shared earlier (`EACCES: permission denied, open '/app/.next/package.json'`) is exactly the kind of issue this approach tries to prevent.

If your container setup has proper permissions configured in other ways, you could potentially remove this step, but it's a common pattern to explicitly manage permissions for directories that need to be written to at runtime.

agennadi: Got it! Feel free to merge the PR when ready

.github/workflows/ci.yml
gamesover: good catch, I used it for testing. reducing frontend `start_period` to `60s`

Issue #262, opened 2025-03-06
Backend container exits prematurely due to healthcheck timing issue


agennadi:
## Context
The backend container exits unexpectedly during the startup process, possibly due to the healthcheck failing too early. The healthcheck currently runs:
`test: [ "CMD-SHELL", "curl -f http://localhost:8000/api/health || exit 1" ]`
However, it seems that `bash /startup.sh` takes too long to complete, leading to premature failures. This causes the CI pipeline to fail, preventing successful deployments.

## Definition of Done
- The backend container remains running and does not exit prematurely.
- CI pipeline runs successfully without exiting due to the healthcheck.
- The healthcheck correctly verifies the service is up and does not trigger failures while startup.sh is still initializing.

## Engineering Details
Relevant files: `compose.yml`, `backend/Dockerfile`, `backend/etl/startup.sh`

PR #264, opened 2025-03-11
Backend healthcheck

agennadi:
Co-authored-by: @gamesover

# Description

Initial approach introduced by @gamesover was to remove the `public` volume and copy the files prom public into the container at build time. However, in this case, data generated by scripts will not persist. The health check passed because there were no ongoing time-consuming file writes.

I suggest changing the order of the commands in the startup.sh script, so that the FastAPI server is available before ETL scripts are completely executed. 
  

#262

## Type of changes
- [X] Bugfix
- [ ] Chore
- [ ] New Feature

## Testing
- [ ] I added automated tests

## How to test

Run CI 

## Clean commits

### DIFF
--- a/backend/etl/startup.sh
+++ b/backend/etl/startup.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 set -e
 
+# Start FastAPI
+uvicorn api.index:app --host 0.0.0.0 --port 8000 &
+
 python backend/database/init_db.py
 python backend/etl/liquefaction_data_handler.py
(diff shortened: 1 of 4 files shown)


backend/etl/startup.sh
```diff
@@ -1,10 +1,13 @@
 #!/bin/bash
 set -e
 
+# Start FastAPI
+uvicorn api.index:app --host 0.0.0.0 --port 8000 &
```
gamesover, line 5: nice

gamesover (approved):
LGTM

PR #301, opened 2025-04-07
cache results from search bar

graample:
this implements a temporary fix that memoizes the first calls made to mapbox using the useCallback React function. we should consider a refactor later down the line, as this is not intended to be a permanent solution.

tested locally using local frontend development build and backend docker containers.

before:
https://github.com/user-attachments/assets/fdfcb466-2e3c-4117-a5c0-6b5e0abcf3ec

after:
https://github.com/user-attachments/assets/6d2cee60-c6e4-4cbe-b86d-78416e6b3514

### DIFF
--- a/app/components/search-bar.jsx
+++ b/app/components/search-bar.jsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useEffect } from "react";
+import { useState, useEffect, useCallback } from "react";
 import { useSearchParams, useRouter } from "next/navigation";
 import {
   HStack,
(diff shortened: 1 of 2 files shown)


nickvisut (approved):
thanks, looks good!

nickvisut:
NOTE: This PR fixed the issue mentioned in this Slack convo https://sfbrigade.slack.com/archives/C07DB8HBV9P/p1744050534052489

possible TODO: create an issue after the fact that references the conversation and problem being solved for reference

PR #446, opened 2025-06-19
Feat/db upserts

MoofSoup:
# Description

Refactors the DataHandler ABC to add insert policy system, enabling child classes to define SQL SET clause for ON CONFLICT DO UPDATE. Child classes can now override `insert_policy()` to return a simple dictionary for conflict resolution, giving a clean API for defining database upserts.

This addresses feedback from the original PR to simplify the over-complicated SQLAlchemy statement approach.

## Type of changes
- [ ] Bugfix
- [x] Chore
- [x] New Feature

## Testing
- [x] I added automated tests

## How to test

Run the existing ETL tests to ensure the refactored insert policy system works correctly:
```bash
# CI runs these tests already
python -m pytest backend/etl/tests/
```

The tests validate that:
- Default behavior (DO NOTHING on conflict) works
- Custom insert policies can be implemented by overriding the method
- Proper error handling and logging occurs

## Clean commits


backend/etl/data_handler.py
agennadi, line 1: Looks great! Thanks for the explicit error handling :D just one nit - unused `Insert` import

backend/etl/soft_story_properties_data_handler.py
agennadi, line 1: You could minimize hardcoding, for example:
```
all_columns = {col.name for col in self.table.__table__.columns}
excluded_fields = {'address', 'update_timestamp', 'sfdata_loaded_at'} # because we use address as PK
conditional_fields = all_columns - excluded_fields
```

PR #454, opened 2025-06-23
docs: add warning about not deleting package-lock.json

### LINKED ISSUE #452: Add note to README to never delete package-lock.json
Add note to README to never delete package-lock.json. Only exception would be for experimentation, but then one must be certain to not check it in by accident or it could cause unintended side effects.

`package-lock.json` helps ensure reproducibility of what's in `node_modules`, so deleting it may disrupt said reproducibility. more concretely, the most dangerous case is deleting both this file and the `node_modules` folder and then running `npm install`; this will mean that NPM relies solely on `package.json` to generate a new `package-lock.json` and thus may diverge greatly from the one that's committed.

also explicitly add a note that in the event of merge conflicts, running `npm install` will "repair" `package-lock.json`, obviating the need to attempt repairing it manually (or, worst case, deleting it). this note is important b/c this is one (misguided) motivation new devs have for deleting the `package-lock.json` file.

NOTE: `package-lock.json` is often ignored during code reviews, so if the above rules aren't followed, a "bad" version of it could still be accidentally merged; to prevent this, it may also be worth looking into ways to prevent a newly generated one (newly generated from whole cloth) from being committed as separate Github issue.

NOTE #2: may also be worth looking into if other package managers have guards against this, or if there are established best practices/tools to prevent a bad commit.

Context:

> turns out that the FOUC was caused by a bad `package-lock.json` wipe in local branch; should probably add a note to README wrt this footgun.

_Originally posted by @nickvisut in https://github.com/sfbrigade/datasci-earthquake/pull/414#discussion_r2160965943_

Hatice-Kocabas:
# Description

Added a detailed note in the README about the importance of not deleting package-lock.json, 
explaining the problems it can cause and how to handle merge conflicts.

Closes #452


nickvisut:
Hi, @Hatice-Kocabas! We appreciate this doc contribution. One request: Can you move this new section underneath the "### Front end" subheading? It's more appropriate there versus under "# Other resources > ## Learn More", which is meant for eg external links.

Hatice-Kocabas:
Thanks for the feedback! I’ve moved the package-lock.json section under the “Front end” heading as suggested. Let me know if there’s anything else I should adjust. 😊

README.md
```diff
@@ -129,6 +129,22 @@ For front end development, first run `npm install`, and then you can run `npm ru
 
 If you need to rebuild the containers, run `npm run docker-back`.
 
+### ⚠️ Please do NOT delete `package-lock.json`
```
nickvisut, line 132:
```suggestion
#### ⚠️ Please do NOT delete `package-lock.json`
```
changing this to an H4 so that it's clear it is part of "### Front end"

nickvisut (changes requested):
Thanks for doing that! I added one tiny code suggestion wrt markdown formatting and then we can merge this.

Issue #439, opened 2025-06-18
Look more carefully at onCoordDataRetrieve parameter types


nickvisut:
![medium](https://www.gstatic.com/codereviewagent/medium-priority.svg)

Consider using a more specific type for the `data` prop in `onCoordDataRetrieve` instead of `any[]` to improve type safety and code clarity. If the structure of the items within these arrays is known, defining a more specific type (e.g., `Feature[]` or `MyCustomDataType[]`) would be beneficial.

```typescript
   onCoordDataRetrieve: (data: {
    softStory: YourSpecificType[] | null;
    tsunami: YourSpecificType[] | null;
    liquefaction: YourSpecificType[] | null;
  }) => void;
```

_Originally posted by @gemini-code-assist[bot] in https://github.com/sfbrigade/datasci-earthquake/pull/437#discussion_r2154112460_

PR #469, opened 2025-06-30
Add HazardData type.

jessemavi:
# Description

- Closes https://github.com/sfbrigade/datasci-earthquake/issues/439

1.  Added a new folder and file for adding TS types
2. Adding a more specific type for the data prop in `onCoordDataRetrieve`. I logged the shape of the values used in the data object(screenshot below). Looks like `exists` will always be returned as a boolean and `last_updated` can be null or a string value.

Note: this is my first PR in the project so definitely possible I missed something.

Screenshot 2025-06-30 at 12 28 38 AM

## Type of changes
- [ ] Bugfix
- [x] Chore
- [ ] New Feature

## Testing
- [ ] I added automated tests

## How to test

## Clean commits

### DIFF
--- a/app/components/home-header.tsx
+++ b/app/components/home-header.tsx
@@ -11,14 +11,19 @@ import Share from "./share";
 import ShareSkeleton from "./share-skeleton";
 import { AnimatePresence, motion } from "framer-motion";
 import SearchBarSkeleton from "./search-bar-skeleton";
-import { CoordinateData } from "../types/index";
+
+type HazardData = {
+  liquefaction: { exists: boolean; last_updated: string | null } | null;
+  softStory: { exists: boolean; last_updated: string | null } | null;
+  tsunami: { exists: boolean; last_updated: string | null } | null;
+};
 
 interface HomeHeaderProps {
   coordinates: number[];
   searchedAddress: string;
   onSearchChange: (coords: number[]) => void;
   onAddressSearch: (address: string) => void;
-  onCoordDataRetrieve: (data: CoordinateData) => void;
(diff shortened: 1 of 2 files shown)


nickvisut (approved):
thank you, @jessemavi! is having a global `types` folder a best practice in a NextJS + TS app, in your opinion?

nickvisut:
> > is having a global types folder a best practice in a NextJS + TS app, in your opinion?
> 
> Thanks for a taking a look and good question @nickvisut!
> 
> Admittedly it was a gemini suggestion after I first put up the PR and I thought it could be a good idea as I have worked in TS apps(not NextJS though) that do separate the types out into another folder.
> 
> But now that I think of it `CoordinateData` doesn't look like a type that is going to be used in multiple places in the app so feels like having it be in a global folder isn't going to actually help much. So I think it might be best to go back to my [initial commit](https://github.com/sfbrigade/datasci-earthquake/pull/469/commits/e3f0cadab0c621b5b03b07b2481c7ebfdfd04dad) and have the type co-located in the component. Maybe better to move it into a separate folder in the future if the need ever arises. Does this sound good to you?

hi @jessemavi, I'm actually fine with either as I don't have a strong preference. Gemini has some good suggestions from time to time.

I do like your suggestion for colocating, though, and perhaps we can do a refactor at some point to see if anything needs to be created/moved to global types.

One other thing I'm realizing is that naming it "CoordinateData" might not be quite the right name since this is technically "hazard data" being passed to an event handler for a UI update (the event being that the coordinate data was just retrieved).

Issue #463, opened 2025-06-26
Remove “risk” language from display cards for Tsunami and Structural Safety.


nickvisut:
**Wooden Soft-Story:** Compliant / Non-Compliant / No data
**Soil Liquefaction:** Low Hazard / High Hazard / Very High Hazard
**Tsunami Hazard Zone:** Out of Zone / In Zone

nickvisut:
cc @agennadi @leela-solomon

agennadi:
cc  @leela-solomon @nickvisut 

This is the risk column in the liquefaction dataset:

```
 liq 
-----
 H
 VH
```
I suggest using "High risk", "Very high risk" and "Low risk". If we want to avoid "risk", we could use "susceptibility" (the official DataSF language).

The risk column in the tsunami dataset:

```
         evacuate         
--------------------------
 Yes, Tsunami Hazard Area
 No, Outside Hazard Area
```

We could use "Hazard area" or "Evacuation zone" (imo, the latter might be misleading - someone might think they need to evacuate now). For "No, Outside hazard area", we can keep "Low risk" or use "Outside hazard area".

The 'status' column in the Soft Stories dataset:

```
           status            
-----------------------------
 Newly Added - Permit Issued
 work Complete, CFC Issued 
 Non-Compliant
 Work Complete, CFC Issued
 Newly Added
```
"Newly Added - Permit Issued" = "Permit Issued"
"Work Complete, CFC Issued" = "Work Complete"
"Non-Compliant" = "Non-Compliant"
"Newly Added" - "Newly Added Soft Story" (?)
The rest could be "No data".

I'd appreciate your thoughts and ideas!

nickvisut:
> cc [@leela-solomon](https://github.com/leela-solomon) [@nickvisut](https://github.com/nickvisut)
> 
> This is the risk column in the liquefaction dataset:
> 
> ```
>  liq 
> -----
>  H
>  VH
> ```
> 
> I suggest using "High risk", "Very high risk" and "Low risk". If we want to avoid "risk", we could use "susceptibility" (the official DataSF language).

If not marked as H or VH, is it implied that there's no data? does the absence of either classification actually mean "low risk"?

> The risk column in the tsunami dataset:
> 
> ```
>          evacuate         
> --------------------------
>  Yes, Tsunami Hazard Area
>  No, Outside Hazard Area
> ```
> 
[...]

PR #501, opened 2025-07-17
Remove “risk” language from display cards

JobeADev:
# Description

Changes to the hazard card's pill labels

#463

## Type of changes
- [ ] Bugfix
- [ ] Chore
- [X] New Feature

## Testing
- [ ] I added automated tests

## How to test

Check text of pills within individual hazard cards when searching addresses

## Clean commits

### DIFF
--- a/app/components/card-hazard.tsx
+++ b/app/components/card-hazard.tsx
@@ -9,6 +9,7 @@ import {
   Spinner,
   Popover,
   Portal,
+  Button,
 } from "@chakra-ui/react";
 import posthog from "posthog-js";
 import Pill from "./pill";
@@ -35,6 +36,8 @@ const CardHazard: React.FC<CardHazardProps> = ({
   showData,
   isHazardDataLoading,
 }) => {
+  // const posthog = usePostHog();
+
   const { title, name, description } = hazard;
   const { exists, last_updated: date } = hazardData || {};
   const labelInfo = PillLabels.filter((label) => label.name === hazard.name)[0];
@@ -53,11 +56,13 @@ const CardHazard: React.FC<CardHazardProps> = ({
   );
 
   const buildHazardCardInfo = () => {
-    return hazard.info.map((infoItem, index) => (
-      <Text as="p" mt="4" key={index}>
-        {infoItem}
-      </Text>
-    ));
+    return (
+      <VStack gap={5} p={5}>
+        {hazard.info.map((infoItem, index) => (
+          <Text key={index}>{infoItem}</Text>
+        ))}
+      </VStack>
+    );
   };
 
   return (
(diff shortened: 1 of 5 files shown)


nickvisut:
hi @JobeADev, thanks for this effort! I'm realizing after discussion w/@agennadi that this can't be fully implemented without backend help (we need more than the "exists" boolean from the API to distinguish between some of the states).

Once we have something additional in place besides the boolean, that will also (unfortunately) mean mapping can't be one-to-one as is done above. We'll likely have to revisit during beta.

There may be an incremental step we can take without backend changes, though. Stay tuned!

nickvisut:
@JobeADev FYI you may want to update your branch and test things due to the library upgrades. also, your Vercel deployment failed. not sure why but you can sign in and check the logs. or, alternatively, if you do another commit and push, another deployment will run.

Also, as discussed, #463 has been updated. Keep in mind that an eventual version of the cards may have pills that don't have exactly two values (or three if you include undefined), so you may want to address that in preparation for that change. Up to you, though!

nickvisut:
hi @JobeADev, FYI your branch name is a little weird (note at top says "The head ref may contain hidden characters: : task/463/Remove-\u201Crisk\u201D-language-from-display-cards") for future reference. I think perhaps b/c stylized quotes are being used. Probably best to stay away from single/double/stylized quotes.

hopefully, it should be fine, though.

app/components/card-hazard.tsx
```diff
@@ -97,34 +100,24 @@ const CardHazard: React.FC<CardHazardProps> = ({
         </Popover.Trigger>
         <Portal>
           <Popover.Positioner>
-            {/* TODO FIXME: can below line be styled with mt={5} width={"348px"} somehow still? how? should it go on `<Popover.Body>`? or elsewhere? */}
-            <Popover.Content>
-              {/* TODO FIXME: can below line replace the <PopoverCloseButton />? */}
+            <Popover.Content maxHeight="unset">
               <Popover.CloseTrigger
```
nickvisut: hi @JobeADev, are these your changes, or changes merged in from `develop`? if the latter, I think you may have to rebase and force push so that the diff only reflects your changes.

app/data/data.js
```diff
@@ -257,3 +257,24 @@ export const TeamMembers = [
     name: "Lola Jackson",
   },
 ];
+
+export const PillLabels = [
+  {
+    name: "softStory",
+    trueData: "Non-Compliant",
+    falseData: "Compliant",
```
nickvisut, line 265: Let's change "Compliant" to "No Data" as per #463's description (context in comments - convo w/@agennadi)

nickvisut: (once API changes land, we can of course update to reflect more specific return value from API)

app/data/data.js
```diff
@@ -257,3 +257,24 @@ export const TeamMembers = [
...
+    noData: "No Data",
+  },
+  {
+    name: "liquefaction",
+    trueData: "High Hazard",
+    falseData: "Low Hazard",
```
nickvisut, line 271: perhaps "Not in Zone" or "Outside of Zone" instead of "Low Hazard"? (until API changes)

nickvisut: (as you suggested)

app/data/data.js
```diff
@@ -257,3 +257,24 @@ export const TeamMembers = [
...
+    noData: "No Data",
+  },
+  {
+    name: "tsunami",
+    trueData: "In Zone",
+    falseData: "Out of Zone",
```
nickvisut, line 277: how about "Not in Zone" etc to match liquefaction?

nickvisut (changes requested):
lgtm aside from some tweaks to the language

app/components/card-hazard.tsx
JobeADev: those are changes merged from develop, I'll give what you suggested a try

JobeADev:
Closing PR in favor of PR #542

Issue #438, opened 2025-06-18
fix useCallback issues


nickvisut:
> ![high](https://camo.githubusercontent.com/6be8dfd86afd9805fdf193065c11418ba1225c487bda10d961f6a486f586fa07/68747470733a2f2f7777772e677374617469632e636f6d2f636f64657265766965776167656e742f686967682d7072696f726974792e737667)
> 
> The `useCallback` hooks for `memoizedGetHazardData` have empty dependency arrays (`[]`). This is problematic because the functions they wrap (`getHazardData`) close over props and state that can change (e.g., `onHazardDataLoading`, `onSearchComplete`, `coordinates` from props).
> 
> To fix this, `getHazardData` should itself be wrapped in `useCallback` with its correct dependencies, or their dependencies should be included in the `useCallback` for `memoizedGetHazardData` if they are to be memoized directly.
> 
> Example for `getHazardData` (if it were directly memoized):
> 
> ```js
> const memoizedGetHazardData = useCallback(async (coords_param = coordinates) => {
>   // ... uses onHazardDataLoading, onSearchComplete, coordinates ...
> }, [onHazardDataLoading, onSearchComplete, coordinates]);
> ```
> 
> Given the `TODO` comment on line 185, this might be a known area for refactoring. However, the current implementation with empty dependency arrays for these complex functions is a high-risk for bugs.

Aware of this, but will save this info for later attempted fixes.

UPDATE: removed the `useCallback` in all branches for `memoizedGetHazardData`
TODO: revisit by making sure `useCallback` calls are passed anonymous functions so that the dependencies can be "seen" and properly identified by the linter and proceed from there.
TODO #2: consider redoing control flow altogether so that the `useEffect` is in a different place and there aren't then two ways to trigger UI updates (both `handleRetrieve()` and the `useEffect()` for URL param changes)

PR #526, opened 2025-07-30
Add useHazardDataFetcher hook and refactored search-bar component to use hook

jessemavi:
# Description

- Related to https://github.com/sfbrigade/datasci-earthquake/issues/438

This PR makes the following changes:
- Moves data fetching logic into a shared hook so the `SearchBar` component focuses more on rendering
- `fetchHazardData` function is memoized within the new hook and is wrapped in `useCallback` with its dependencies to ensure the function always has access to the latest values, preventing it from becoming "stale."
- Adds unit tests for the `useHazardDataFetcher` hook
- Updates jest config to map the `@` alias to the app directory when running tests. `@` is used when importing ui components like `Toaster`

I think this should fix the issue linked above but not 100% sure. It should at least be an improvement to separate the data fetching logic.

Screen recording shows that existing behavior hasn't changed and we are fetching hazard data successfully.

https://github.com/user-attachments/assets/5711fafe-b794-4a04-855a-b6197b77a041

## Type of changes
- [ ] Bugfix
- [x] Chore
- [ ] New Feature

## Testing
- [x] I added automated tests

## How to test
- Select an address and see that the behavior is unchanged and we fetch hazard data and display the results
- Run tests using `npm test`

## Clean commits

### DIFF
--- a/app/components/__tests__/hooks/useHazardDataFetcher.test.tsx
+++ b/app/components/__tests__/hooks/useHazardDataFetcher.test.tsx
@@ -1,5 +1,5 @@
 import { renderHook, act } from "@testing-library/react";
-import { useHazardDataFetcher } from "../../hooks/useHazardDataFetcher";
+import { useHazardDataFetcher } from "../../../hooks/useHazardDataFetcher";
 import { toaster } from "@/components/ui/toaster";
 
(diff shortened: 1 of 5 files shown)


jvaneyken (approved):
LGTM

Issue #525, opened 2025-07-29
Add Very High Risk and High Risk to Soil Liquefaction Backend


AdamFinkle:
**Is your feature request related to a problem? Please describe.**
To support the Very High Susceptibility Risk and High Susceptibility information display on the frontend, the backend must supply this information.

**Describe the solution you'd like**
`liquefaction_api.py`
- `get_liquefaction_zones`: instead of returning one `LiquefactionFeatureCollection`, return one for High Susceptibility and another for Very High Susceptibility
- `is_in_liquefaction_zone`: replace with `in_high_susceptibility_liquefaction_zone` and `in_very_high_susceptibility_liquefaction_zone`
- This change will require corresponding change in `API_ENDPOINTS` in `endpoints.ts`
- [x] `liquefaction_zones.py`: add a `susceptibility_level` field to `LiquefactionZone`
- ~~`liquefaction_schemas.py`: add a `susceptibility_level` field to `LiquefactionProperties` and the properties field of the return of `LiquefactionFeature.from_sqlalchemy_model`~~

**Describe alternatives you've considered**
There is no alternative because the frontend needs this information.

qredwoods:
@NewtonLC and I spent some of today's Hack Night reviewing this issue and the history. Newton will take a crack at it with my help as needed and we'll ask for further support. FYI our current understanding is that the functions is in_high_susceptibility_liquefaction_zone and in_very_high_susceptibility_liquefaction_zone would be similar to is_in_liquefaction_zone, with differences analogous to the differences between get_liquefaction_zones and get_very_high_susceptibility. Next step is Newton will study the liquefaction_api.py file more closely and attempt a draft of these new functions. More established folks feel free to step in / advise.

One question is if front end needs to retain this being returned as a boolean like is currently the case with exists, or if perhaps a severity data item could exist instead with something like 0 is outside of liq zone, 1 is high, and 2 is very high.

NewtonLC:
From looking into the is_in_liquefaction_zone, it seems like the function already returns the zone's susceptibility level(the `liq` attribute), and returns it in the InLiquefactionZoneView object returned by the function. So, we can do:
`exists == False` --> 'Not in Zone'
`exists == True and liq == 'H'` --> 'High Hazard'
`exists == True and liq == 'VH'` --> 'Very High Hazard'
as outlined in issue #463.

[...]

PR #539, opened 2025-08-02
Split soil liquefaction api and tests into high susceptibility and very high susceptibility

AdamFinkle:
# Description

The soil liquefaction API endpoint delivers all soil liquefaction multipolygons as one collection of features and has a test to determine whether a point lies among them.  This change splits the first endpoint into one for high-susceptibility areas and another for very-high-susceptibility areas and the second endpoint into one that checks whether a point is in a high-susceptibility area; another, very-high.

Note that this pull request should not be merged without commits from the frontend team to adapt the frontend to display soil liquefaction zones in high-and very-high-susceptibility areas because even though the frontend tests pass, the app will not work unless the frontend is adapted.

#525 

## Type of changes
- [ ] Bugfix
- [ ] Chore
- [x] New Feature

## Testing
- [x] I added automated tests

## How to test

Run the automated backend tests.

## Clean commits


agennadi (changes requested):
1. LiquefactionFeatureCollectionResponse is not a valid geojson and won't be parsed by the map. There is no need to combine these two geojsons together. I suggest using three distinct endpoint: keep "/" for all risk zones as we had it (for backward compatibility), "/high-risk" and "/very-high-risk". Each of them should return a single FeatureCollection. If you prefer to keep them under one endpoint, you could use a query parameter.
2. Two separate endpoints `/in_very_high_susceptibility_liquefaction_zone` and `/in_high_susceptibility_liquefaction_zone` don't have a practical use case. Making two api calls per point is inefficient. Instead, add the risk status field to `IsInLiquefactionZoneView`, keep the original `is_in_liquefaction_zone` method and return the risk status along with "exist" and "last_updated".

I haven't run the code or reviewed it carefully yet, but these two issues need to be addressed first.

AdamFinkle:
@agennadi I have made the changes you've requested.

agennadi (approved):
Looks good!

agennadi (changes requested):
is-in-liquefaction-zone always returns "low risk" for all hazards

agennadi (approved):

Issue #391, opened 2025-05-28
Add ability to toggle data layers and a map legend


juliacordero:
**Is your feature request related to a problem? Please describe.**
Users were confused by the visualization of each data set on the initial map.

**Describe the solution you'd like**
Add a map legend with ability to toggle each data layer: soft story, liquefaction, tsunami risks.

nickvisut:
hi @jocelyn-su, @JobeADev created a POC that can be seen at the preview deployment at #546. Can you please take a look?

PR #546, opened 2025-08-06
feature/391/add-toggle-data-layers

JobeADev:
# Description

Makes map legends, and data layers of map, toggleable.

#391

## Type of changes
- [ ] Bugfix
- [ ] Chore
- [X] New Feature

## Testing
- [ ] I added automated tests

## How to test

Click on map legends to toggle appearance of data layers on map.

## Clean commits


app/components/map.tsx
```diff
@@ -26,6 +28,64 @@ const Map: React.FC<MapProps> = ({
...
+              "circle-color": "#A0AEC0", // gray/400
+            },
+          });
+        } else {
+          map.removeLayer("softStoriesLayer");
+        }
+      }
+    }
+    return;
+  };
```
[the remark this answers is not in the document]
JobeADev: done

app/components/map.tsx
```diff
@@ -164,6 +180,10 @@ const Map: React.FC<MapProps> = ({
     }
   }, [coordinates, liquefactionData, softStoryData, tsunamiData]);
```
nickvisut, line 181: Can we add a known issue about the "de-sync" discussed for the early toggle edge case, @JobeADev? (separately, I wonder if adding the booleans to deps with default "false" would get it working)

app/components/report-hazards.tsx
```diff
@@ -1,19 +1,46 @@
...
 const ReportHazards = ({
   addressHazardData,
   isHazardDataLoading,
+  setLayerToggleObj,
 }: {
   addressHazardData: HazardData;
   isHazardDataLoading: boolean;
+  setLayerToggleObj: Dispatch<SetStateAction<LayerToggleObjProps>>;
 }) => {
+  const [toggledStates, setToggledStates] = useState<ToggledStatesProps>([]);
```
nickvisut, line 22: As discussed, it may make sense to lift this state up one level in a future PR.

app/data/data.js
```diff
@@ -226,6 +226,16 @@ export const TeamMembers = [
     role: "Front End Engineer",
     name: "Josh van Eyken",
   },
+  {
+    id: 21,
+    role: "Front End Engineer",
+    name: "Jobe Abdellah",
```
nickvisut, line 232: 🎉

app/components/legend-clicked-context.tsx
```diff
@@ -0,0 +1,80 @@
+"use client";
```
nickvisut: Can you add a comment to explain this file for future devs? It would be good if it was made clear that it's not currently being imported etc.

Perhaps it could be moved to some sort of example/snippet folder.

app/components/address-mapper.tsx
```diff
@@ -89,13 +98,15 @@ const AddressMapper: React.FC<AddressMapperProps> = ({
             <ReportHazards
               addressHazardData={addressHazardData}
               isHazardDataLoading={isHazardDataLoading}
+              setLayerToggleObj={setLayerToggleObj}
```
nickvisut, line 101: Nit: I've come to understand recently that it's better to pass a callback containing a setter call vs directly passing a setter like this.


We already do this on lines 91-93, though, so don't sweat it. It can probably be looked into when we tackle all similar instances.

app/components/report-hazards.tsx
```diff
@@ -35,16 +62,40 @@ const ReportHazards = ({
             Legend:
           </Text>
           <KeyElem name="Soft story" color="grey.400" icon={<FaCircle />} />
+          <Switch.Root
```
nickvisut, line 70: Please add a follow-up issue or comment for refactoring this section into a loop/map as per your suggestion.

app/components/map.tsx
```diff
@@ -13,20 +14,35 @@ interface MapProps {
...
+  const handleToggleLayers = () => {
+    if (!mapRef.current) return;
+    const map = mapRef.current;
+
+    const layerId = layerToggleObj.layerId;
+
+    if (layerId && map.getLayer(layerId)) {
+      const newVisibility = layerToggleObj.toggleState ? "visible" : "none";
+      // sets new visibility property value for layer, creating the "toggling" effect
+      map.setLayoutProperty(layerId, "visibility", newVisibility);
```
nickvisut, line 42: Ah, the secret sauce! 😌

app/components/map.tsx
```diff
@@ -78,7 +94,7 @@ const Map: React.FC<MapProps> = ({
       map.touchZoomRotate.disableRotation(); // turn off rotate w/touch
 
       const nav = new mapboxgl.NavigationControl({ showCompass: false });
-      map.addControl(nav, "right");
+      map.addControl(nav, "bottom-right");
```
nickvisut, line 97: Thanks for the bonus!

app/components/map.tsx
```diff
@@ -164,6 +180,10 @@ const Map: React.FC<MapProps> = ({
     }
   }, [coordinates, liquefactionData, softStoryData, tsunamiData]);
 
+  useEffect(() => {
```
nickvisut, line 183: Good thinking isolating this into its own useEffect.

app/data/data.js
```diff
@@ -278,3 +288,5 @@ export const PillData = [
     noData: "No Data",
   },
 ];
+
+export const LayerIds = ["softStoriesLayer", "seismicLayer", "tsunamiLayer"];
```
nickvisut, line 292: Thanks, this'll make it easier to add new layers!

nickvisut (approved):
Thanks for the work on this! Looks great.

I included one significant request wrt to the loose context file and some minor ones wrt to adding the follow-up Issues and/or Comments that we discussed.

There is also a nit and commentary that likely don't need any follow-up at all.

Issue #508, opened 2025-07-24
Simplify UI loading skeletons


nickvisut:
## Context
Right now, we have a couple of fallback skeletons for usage w/React's `Suspense`. Their filenames end with -skeleton.tsx. Code is repeated between the components they're meant for and the skeletons themselves since the skeletons are eseentially stripped-down versions of the components.

These loading skeletons were written in haste during React/Next upgrades to satisfy new library requirements while also keeping feature parity,. However, the resulting repeated code makes maintenance difficult.

Some possible ways to address this:
- use Chakra UI v3's 'Skeleton' component, which creates a grey box of specified width and height as a loading effect instead
- componentize the repeated parts
- other option that follows best practices of Suspense's `fallback` prop

## Definition of Done
The repeated code is reduced or even, ideally, eliminated.

## Engineering Details
Look for any mentions of `<Suspense>` with a `fallback` prop. Compare the component wrapped by the `Suspense` to the skeleton component referred to by the fallback prop. Note that the skeleton file looks like a leaner version of the component that is wrapped within the Suspense. Suggest looking into best practices for Suspense fallbacks and following those if they differ from our approach.

jvaneyken:
In Progress PR needs to be revisited to emphasize a maintainable solution.

nickvisut:
Need to revisit how we approach this; editing the UI components shouldn't mean also having to edit the related fallback components (ideally).

PR #558, opened 2025-08-14
Fix duplicate hazard data requests when selecting an address

### LINKED ISSUE #547: Fix multiple duplicate hazard data requests when selecting an address
## Context
It looks like we make duplicate requests to fetch hazard data when selecting an address. On the first address selected after initial page load we make 3 identical requests for each hazard type to the API to get hazard data. On subsequent address selections we make 2 identical requests for each hazard type to the API. Not sure how much additional resources this is costing but we definitely don't want to make multiple requests for no reason.

Screenshot shows multiple requests for a single address(3 for each hazard so 9 total) in dev but confirmed this is also happening in prod.

Image

This is likely also leading to the "flashing" behavior seen below in the `CardHazard` Pill when selecting an address.

https://github.com/user-attachments/assets/6e6fce1a-cda7-4aa7-af41-f792384c37cc

## Definition of Done
Remove the duplicate requests so only 1 request is done for each hazard.

## Engineering Details
Its possible that there could be some unnecessary re-renders that are happening and those are leading to the duplicate data fetches but more investigation is needed. Will need to look more into component life-cycle events related to mounting, rendering and unmounting.

jessemavi:
# Description

- Resolves https://github.com/sfbrigade/datasci-earthquake/issues/547

This PR does the following:
- Moves the state management/handling of url params, `address` and `coordinates` and data fetching to the `AddressMapper` component and out of the `SearchBar` component which gets rid of the multiple data fetches coming from re-renders when selecting an address. Now the url params are being updated in `AddressMapper` and that change will lead to a `useEffect` being called to update state.
- Update `HomeHeader` and `SearchBar` components to be more presentational and pass down only one function `handleSearchChange` as a prop to call when selecting an address.
- Add tests for `AddressMapper`.
- Minor naming updates to `HazardDataFetcher` hook and its tests.  This is because the hook is now being called in `AddressMapper` instead of `SearchBar`.

See Gemini AI summary below for more info.

Screen recording below shows only one data fetch for each of the three hazards instead of multiple for the following user actions:
1. Selecting an address after initial page load
2. Navigating to another address
3. Using back and forward buttons
4. Refreshing the page

https://github.com/user-attachments/assets/2063f359-dd5e-404a-b381-6ad1bd7c4244

## Type of changes
- [ ] Bugfix
- [x] Chore
- [ ] New Feature

## Testing
- [x] I added automated tests

## How to test

## Clean commits


nickvisut:
hi @jessemavi, you ought to have access to the Vercel credentials now, hopefully. Check for an email from Google.

In the meanwhile, here's the error message that shows up in Vercel:
```
Failed to compile.
./app/components-test-lib/search-bar-client-wrapper.tsx:12:7
Type error: Type '{ coordinates: number[]; onSearchChange: () => void; onAddressSearch: () => void; onCoordDataRetrieve: () => void; onHazardDataLoading: () => void; onSearchComplete: () => void; }' is not assignable to type 'IntrinsicAttributes & SearchBarProps'.
  Property 'coordinates' does not exist on type 'IntrinsicAttributes & SearchBarProps'.
  10 |     {/* NOTE: This Suspense boundary is being used around a component that utilizes `useSearchParams()` to prevent entire page from deopting into client-side rendering (CSR) bailout as per https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout */}
  11 |     <SearchBar
> 12 |       coordinates={[0, 0]}
     |       ^
  13 |       onSearchChange={() => {}}
  14 |       onAddressSearch={() => {}}
  15 |       onCoordDataRetrieve={() => {}}
Next.js build worker exited with code: 1 and signal: null
Error: Command "npm run build" exited with 1
```

I believe this error message can also be triggered locally by running `npm run build` locally. If so, adding this as a useful step/developer tip to the `README` would be appreciated 🙏 .

It also reminds me that we probably ought to add `npm run build` (`next build`) to our GitHub CI pipeline so that both Vercel and GitHub are [doing the same thing](https://vercel.com/docs/builds/configure-a-build#build-command).

jessemavi:
@nickvisut thanks for sharing the error message! Pushed up some updates and marked as ready to review. I actually could have seen this error just by looking at my file explorer and seen that this file was in red and had errors but missed it.

`npm run build` is definitely helpful to know and yes I can add some info about it to the README. Running this also helped because it showed me the error for `useSearchParams() should be wrapped in a suspense boundary` so needed to update that for the `AddressMapper` component since it is now using `searchParams` and it also shows any eslint errors so helpful if you don't have eslint installed and can fix these before pushing up changes. I didn't get an email from Vercel though I didn't ever create an account so maybe need to do that. But if I can just do `npm run build` and see if a build is successful and what the errors are then that's all I really need for now.

app/components-test-lib/search-bar-client-wrapper.tsx
```diff
@@ -8,14 +8,7 @@ import SearchBarSkeleton from "../components/search-bar-skeleton";
...
     {/* NOTE: This Suspense boundary is being used around a component that utilizes `useSearchParams()` to prevent entire page from deopting into client-side rendering (CSR) bailout as per https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout */}
-    <SearchBar
-      coordinates={[0, 0]}
-      onSearchChange={() => {}}
-      onAddressSearch={() => {}}
-      onCoordDataRetrieve={() => {}}
-      onHazardDataLoading={() => {}}
-      onSearchComplete={() => {}}
-    />
+    <SearchBar onSearchChange={() => {}} />
```
jessemavi: Because this PR removes `useSearchParams` from the `SearchBar` component I don't believe we need to continue wrapping it in a Suspense boundary. But thought I'd leave it since there is a `SearchBarSkeleton` fallback that was already created and being used here. I can remove the comment above though on why it is being wrapped. Or if we want to remove the Suspense boundary and no longer use the fallback I can look into doing that instead.

app/components/search-bar.tsx
```diff
@@ -33,34 +23,16 @@ const autofillOptions: AddressAutofillOptions = {
 // NOTE: UI changes to this page ought to be reflected in its suspense skeleton `search-bar-skeleton.tsx` and vice versa
 // TODO: isolate the usage of `useSearchParams()` so that the Suspense boundary can be even more narrow if possible
 interface SearchBarProps {
-  coordinates: number[];
-  onSearchChange: (coords: number[]) => void;
-  onAddressSearch: (address: string) => void;
-  onCoordDataRetrieve: (data: HazardData) => void;
-  onHazardDataLoading: (hazardDataLoading: boolean) => void;
-  onSearchComplete: (searchComplete: boolean) => void;
+  onSearchChange: (coords: number[], address: string) => void;
```
nickvisut, line 26: this is a lot cleaner, thanks!

nickvisut:
Hi @jessemavi, looks great so far. May take a while to review b/c have to deep dive and understand both the code and its tests.

What might help is if you can elaborate on how the solution works. More specifically, you write the following:
> Moves the state management/handling of url params, address and coordinates and data fetching to the AddressMapper component and out of the SearchBar component which gets rid of the multiple data fetches coming from re-renders when selecting an address

This is helpful. Do you mind describing--in a bit more detail--the root cause of the multiple calls and how it's addressed?

I'm asking so that I can grok it more easily as a code reviewer. TIA!

app/components-test-lib/search-bar-client-wrapper.tsx
nickvisut: You're probably right that it's no longer needed here nor is the skeleton if the `useSearchParams()` has been moved. However, I believe that each instance of `Suspense` ideally has a skeleton fallback that initially shows while waiting, so I think we need a fallback for `AddressMapper`. I could be wrong, though, so feel free to correct me. We do currently have fallbacks for two or three Suspenses, and we're missing a fallback for one Suspense (which I put a comment on below).

If we do need a fallback, then please see the following issue and PR for context:
- https://github.com/sfbrigade/datasci-earthquake/issues/508 (describes that our current fallback skeletons may not follow best practices and require too much maintenance; suggests trying Chakra's `Skeleton` component instead)
- https://github.com/sfbrigade/datasci-earthquake/pull/524 (PR that attempts to replace the home skeleton's `loading.tsx`, which is basically a Next pattern for a `Suspense` fallback that auto-wraps an entire page; although it "looks" good, it's written in such a way that may actually increase maintenance)

Having said that, please take a look at [this other comment](https://github.com/sfbrigade/datasci-earthquake/pull/558#discussion_r2296862338) wrt the effect of `Suspense` first.

app/components/address-mapper.tsx
```diff
@@ -40,19 +43,101 @@ const AddressMapper: React.FC<AddressMapperProps> = ({
   tsunamiData,
   liquefactionData,
 }) => {
-  const [coordinates, setCoordinates] = useState(defaultCoords);
-  const [searchedAddress, setSearchedAddress] = useState("");
+  const searchParams = useSearchParams();
```
nickvisut, line 46: please see [comment above](https://github.com/sfbrigade/datasci-earthquake/pull/558/files#r2296854453a)

app/components/search-bar.tsx
```diff
@@ -133,33 +76,6 @@ const SearchBar = ({
...
-  }, [
-    searchParams,
-    memoizedOnAddressSearch,
-    memoizedOnSearchChange,
-    memoizedUpdateHazardData,
-  ]);
-
   return (
     <chakra.form position={"relative"} onSubmit={onSubmit}>
       <Suspense>
```
nickvisut, line 81: FYI we are already missing a skeleton/fallback here as an existing example of how a missing fallback may cause issues

app/page.tsx
```diff
@@ -40,11 +41,13 @@ const Home = async () => {
   }
   return (
     <Flex direction="column">
-      <AddressMapper
-        softStoryData={softStoryData}
-        tsunamiData={tsunamiData}
-        liquefactionData={liquefactionData}
-      />
+      <Suspense>
```
nickvisut: please see [comment above](https://github.com/sfbrigade/datasci-earthquake/pull/558/files#r2296854453a)

nickvisut: In addition to the comment above about missing `fallback` prop, I have a few follow-up questions. Keep in mind that I haven’t done deep research into the consequences of using Suspense and best practices around it. These don’t necessarily require action except for the first question. I did want to at least get your thoughts on them, though:

Does moving the Suspense boundary to wrap the AddressMapper component have an adverse effect on the way the app renders? If you could research the answer, and perhaps even test old vs new (e.g., with with throttling/slow connection emulation in the dev tools), that could give us the answer we need.
  - if not, then great! no further action needed besides removing the old Suspense (around Search Bar) and relevant skeleton
  - if so, is there a way to somehow limit/isolate the scope of the Suspense to something smaller than AddressMapper whlie keeping your changes intact? I’m guessing the Suspense is required around the whole thing b/c it seems `useSearchParams()` isn’t available until late in the component lifecycle, but maybe I’m missing something
    - if so, can you attempt to isolate it in this PR? 
    - if not, what if we relied on server-side routing for the query parameters instead? I played with this on an alt branch, but didn’t get the chance to fully test or benchmark it, nor did I contrast how much compute (I believe this is captured [in Vercel Functions stats](https://vercel.com/blog/understanding-vercel-functions)) it takes up vs our existing client-side query parameter handling

jessemavi:
> This is helpful. Do you mind describing--in a bit more detail--the root cause of the multiple calls and how it's addressed?

Hey @nickvisut , sure can explain in more detail. The root cause comes from currently having two places in `SearchBar` where we fetch hazard data: 1) in the `handleRetrieve` function when selecting an address suggestion and 2) the useEffect that calls `memoizedUpdateHazardData`. In the screenshot below after selecting a suggestion, you can see the console logs from the first data fetch inside `handleRetrieve` and then the second and third data fetches from the useEffect.

The changes in this PR are a combination of fixing the multiple data fetches and also some refactoring:

- Centralized state: The `AddressMapper` component now holds all the relevant state for hazard data, including coordinates, searchedAddress, and addressHazardData. This reduces the props sent down to `SearchBar` to only `onSearchChange`.

- URL as the source of truth: The `SearchBar` component no longer directly fetches data. Its `onSearchChange` prop now only updates the URL with the new lat, lon, and address parameters.

- useEffect reacts to URL changes: The useEffect hook in `AddressMapper` is the only place where the `updateHazardData` function is called. It listens for changes to the `searchParams` object so when the URL is updated (either by the SearchBar or by the user navigating directly to a URL), the hook does a single data fetch.

- Preventing redundant renders with useRef: The `coordinatesRef` stores the last fetched coordinates. The useEffect hook checks if the URL's coordinates are different from the ones in the ref before triggering a new fetch. This prevents an infinite loop that could occur since coordinates is in the dependency array of the useEffect hook.

Screenshot 2025-08-26 at 12 49 48 PM

jessemavi:
> Does moving the Suspense boundary to wrap the AddressMapper component have an adverse effect on the way the app renders? If you could research the answer, and perhaps even test old vs new (e.g., with with throttling/slow connection emulation in the dev tools), that could give us the answer we need.

@nickvisut  Good question! From looking into it I don't believe so. Tested it throttling with and without wrapping `AddressMapper` in Suspense and didn't notice a difference in the component rendering. In our case we're using the Suspense boundary to handle the client-side hydration of a component that uses a client-only `useSearchParams` hook and I think the time for this to happen is so short that it's not even really noticeable to the user.

I'm not sure if having a fallback is all that helpful in our case where we're just adding it because of `useSearchParams` as opposed to adding it because of data fetching, lazy loading or some sort of complex synchronous calculation. So yes think removing the one around `SearchBar` could be good.

One change I think I should make to the Suspense component is adding the fallback prop and just setting it to null instead of not having the prop at all. `<Suspense fallback={null}>` Looks like this is how it's done in other codebases like in the [React library](https://github.com/search?q=repo%3Afacebook%2Freact+%3CSuspense+fallback%3D%7Bnull%7D%3E&type=code) where they set it to null and use `<Suspense>` occasionally but only in tests.

jessemavi:
@nickvisut @JobeADev pushed up some updates and this is ready to be reviewed/approved when you get a chance.

I looked into the useEffect and seeing if I could do it in an alternative way without the ref. It looks like it "works" without the ref and without the comparison of new vs. old coords but the ref prevents the component from re-fetching data and updating state if/when a re-render of `AddressMapper` occurs without a change in the URL's coordinates. Doesn't look like it's happening now but using the ref and doing the comparison guards against it and is a more robust solution. Using a ref is fairly common in certain situations and think it works well here for a value(coordinates) that should trigger a side effect(API call) only when they change. Can come back to this if I know that `AddressMapper` isn't ever going to be re-rendered in a case other than when the url params change and can refactor to make it simpler.

I was able to get rid of `coordinates` from useEffect dependency array and just use the current `coordinatesRef` value instead in the `else if` statement so did improve it and now it should only run when `searchParams` changes.

I also made the following minor changes:
- removed the `Suspense` around `SearchBar` as discussed in one of the resolved comments above since it no longer uses `searchParams` and set the fallback to null for the `Suspense` around `AddressMapper`
- in `AddressMapper` changed `searchedAddress` state default value to `null` instead of an empty string. Since this value is coming from an API response I think null could better for the initial value as opposed to it being something else like a controlled form input where an empty string is better.

jvaneyken (approved):
LGTM

PR #608, opened 2025-08-27
updated .env.example

agennadi:
# Description

PRs from forks break because `.env.example` doesn't have dummy values for the most recent variables. I added the variables to the file. 

## Type of changes
- (X) Bugfix
- ( ) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests

## Clean commits


nickvisut (approved):
good to know, and thanks for the fix! is there a way to have the Action somehow auto-update the example file in the future?

agennadi:
> good to know, and thanks for the fix! is there a way to have the Action somehow auto-update the example file in the future?

Good idea! I will look into that

PR #626, opened 2025-09-04
chore/624/adjust-desktop-map

### LINKED ISSUE #624: Desktop & Laptop map adjustments
## Context
see #581

## Definition of Done
Desktop & Laptop view of map matches Figma example

## Engineering Details
see #581

JobeADev:
# Description

Capture9

Map adjusted to match Figma -> https://www.figma.com/design/2yjQGc0lCYCbnfq5qotgEG/SafeHome-Design?node-id=1278-3050&p=f&t=HSkyxCrvACSrE5i8-0
also re-enabled default scrolling on map to zoom, preventing the "ctrl + scroll" text from appearing.

closes #624 

## Type of changes
- ( ) Bugfix
- (X) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests

## How to test

open latest deployment, and scroll on map.

## Clean commits


app/components/map.tsx
```diff
@@ -78,8 +78,8 @@ const Map: React.FC<MapProps> = ({
         dragRotate: false, // turn off rotation on drag
         touchPitch: false, // turn off pitch change w/touch
         touchZoomRotate: true, // turn on zoom/rotate w/touch
-        keyboard: true, // turn on keyboard shortcuts
-        cooperativeGestures: true, // scroll-to-zoom requires using the control or command key while scrolling to zoom the map
+        // keyboard: true, // turn on keyboard shortcuts
+        // cooperativeGestures: true, // scroll-to-zoom requires using the control or command key while scrolling to zoom the map
```
jvaneyken: what are the defaults? Should we delete these lines or set them to false instead of commenting them out?

JobeADev: I think both may be false by default, but you're right I'll delete them.

app/components/address-mapper.tsx
```diff
@@ -176,7 +176,15 @@ const AddressMapper: React.FC<AddressMapperProps> = ({
         isSearchComplete={isSearchComplete}
         onSearchChange={handleSearchChange}
       />
-      <Box w="full" h={{ base: "1400px", md: "1000px" }} m="auto">
+      <Box
```
nickvisut, line 179: as discussed, this is a great step in the right direction for allowing the user to more easily scroll down AND interact w/the map for common desktop use cases, so we ought to probably get it merged in ASAP

next step would be to have the height depend on the viewport height (probably future PR)

app/components/card-hazard.tsx
```diff
@@ -77,24 +77,20 @@ const CardHazard: React.FC<CardHazardProps> = ({
   };
 
   const handleSwitchClick = (num: number, checked: boolean) => {
-    const newArray = [];
-    const obj = {
+    const newArray = [...toggledStates];
```
nickvisut, line 80: nice refactor!

app/components/card-hazard.tsx
```diff
@@ -77,24 +77,20 @@ const CardHazard: React.FC<CardHazardProps> = ({
...
-    setLayerToggleObj(obj);
+    });
   };
 
   return (
     <Card.Root
       flex={1}
-      maxW={{ base: 336 }}
-      minH={{ base: 184 }}
+      maxW={{ base: 320, "2xl": 336 }}
```
nickvisut, line 92: is this tweak a UX request?

app/components/card-hazard.tsx
```diff
@@ -137,7 +133,14 @@ const CardHazard: React.FC<CardHazardProps> = ({
...
           </Card.Header>
           <Card.Body textAlign="left" p={0} mb={"6px"}>
-            <Text textStyle="textMedium" layerStyle="text">
+            <Text
+              textStyle="textMedium"
+              layerStyle="text"
+              fontSize={{
+                base: description.length >= 105 ? 14.4 : 15.2,
+                "2xl": description.length >= 105 ? 15.2 : 16,
+              }} // TODO: find better solution such as adjusting description text or whitespace between characters
```
nickvisut: what is this bit of logic solving? separately, is there a way to stick to the font sizes in the theme?

app/components/search-bar.tsx
```diff
@@ -57,7 +57,9 @@ const SearchBar = ({ onSearchChange }: SearchBarProps) => {
   const handleAddressChange = (event: ChangeEvent<HTMLInputElement>) => {
     setInputAddress(event.currentTarget.value);
     // shows hint again upon further search param changes without selection of suggestion
-    if (!suggestionSelected) {
+    if (!suggestionSelected) setSuggestionsAvailable(false);
```
nickvisut, line 60: can't quite wrap my head around the logic here; mind explaining it?

nickvisut (commented):
looks great! added a few comments and questions.

app/components/card-hazard.tsx
JobeADev: No that was my own choice, to make sure there would be sufficient white-space for laptops, while still being proportionate.

app/components/card-hazard.tsx
JobeADev: This is to prevent the text of the third card from wrapping more, causing that card's height to change and be larger than the other two. I could try refactoring with the theme in mind,

app/components/search-bar.tsx
JobeADev: Its to allow the hint to reappear if the user manually (using the backspace) clears the search bar rather then using the x towards the right-hand side to clear. Previously, if the user cleared manually, there was no logic in place to reset `suggestionSelected` to false. This is meant to cause that reset after an address search if the user clears the characters of the search bar to the length in the statement (`inputAddress.length <= 3`). It was something I sort of just threw together after noticing the issue, so if a refactor would be desired, I totally understand!

app/globals.css
```diff
@@ -7,3 +7,12 @@
...
   border-radius: 0;
 }
+
+.mapboxgl-ctrl-group button {
+  width: 40px;
+  height: 40px;
+}
+
+.mapboxgl-ctrl-bottom-right {
+  margin-right: 16px;
```
nickvisut, line 17: I checked the DOM and it looks like you have to do this bc of the attribution/copyright in the lower-right corner?

nickvisut (approved):
LGTM. I left one question to confirm something, but feel free to merge.

app/globals.css
JobeADev: No necessarily, I believe the attribution/copyright is affected by the same css properties as control elements in the bottom right. To my memory, they sort of bundled these elements together like that, which is odd. As far the use of the globals.css, from what I found in my research, that was the best method as the other alternative would be using selectors in the useEffect and change the properties there. Something like this for example -> 
```
         const navButtons = document.querySelectorAll(
           ".mapboxgl-ctrl-group button"
         );

         // Loop through and apply inline styles to each button
         navButtons.forEach((button) => {
           button.style.width = "40px";
           button.style.height = "40px";
         });
```

Issue #648, opened 2025-09-22
Wrong statuses are displayed for retrofitted and newly added soft stories


agennadi:
**Describe the bug**
- Report cards for soft stories with status "Work Complete, CFC Issued" display "No data" instead of "Compliant".
- Report cards for soft stories with status "Newly Added - Permit Issued" return "Non-compliant" instead of "No data"

**To Reproduce**
1. Go to 'develop.safehome.report'
2. Search 510 24TH AV or 2439 LARKIN ST (CFC is issued for these addresses)
3. The report shows "No data" instead of "Compliant"
4. Search 1254 48TH AV ("Newly Added")
5. The report shows "Non-Compliant" instead of "No data" 
6. Compare with safehome.report

**Additional context**
Logic in `is_soft_story()` in `backend/api/routers/soft_story_api.py` should be rewritten:
"Non-Compliant" returns `{exists:true}`
"Work Complete, CFC Issued" returns `{exists:false}`
All other statuses return {`exists: None}`

`data.js:`
```
export const PillData = [
  {
    name: "softStory",
    trueData: "Non-Compliant",
    falseData: "No Data", // I believe this should be "Compliant"
    noData: "No Data",
  },
```

PR #649, opened 2025-09-22
Fixed statuses for retrofitted and newly added soft stories

agennadi:
# Description

1. Soft Story API distinguishes between "CFC issued", "non-compliant" and all other statuses.
2. ETL workflow is updated to generate the soft story geojson only for non-compliant soft stories.
3. PillData is updated for `falseData`.

#648

## Type of changes
- (X) Bugfix
- ( ) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests

## How to test
1. Search 510 24TH AV or 2439 LARKIN ST -> the report should return "Compliant".
4. Search 1254 48TH AV -> the report should return "No Data"

## Clean commits


backend/api/routers/soft_story_api.py
```diff
@@ -107,21 +108,22 @@ def is_soft_story(
...
 
-        exists = property is not None
-        last_updated = property.update_timestamp if property else None
+        if property:
+            last_updated = property.update_timestamp
+            status_lower = property.status.lower()
+            if status_lower == STATUS_WORK_COMPLETE_LOWERCASE:
+                exists = False
+            elif status_lower == STATUS_NON_COMPLIANT:
+                exists = True
```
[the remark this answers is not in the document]
agennadi: Current code is ok, there is no need to add more status mappings.

backend/etl/soft_story_properties_data_handler.py
```diff
@@ -77,7 +77,7 @@ def _convert_to_geojson(self, data: list[dict]) -> dict:
         for item in data:
             status = item["status"]
             wkt_point = item.get("point")  # Extract 'point' value (WKT)
-            if status and status.lower() == "work complete, cfc issued":
+            if status and status != "Non-Compliant":
```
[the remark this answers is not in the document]
agennadi: Addressed

backend/api/routers/soft_story_api.py
```diff
@@ -107,21 +108,22 @@ def is_soft_story(
...
     try:
+        exists = None
+        last_updated = None
         point = from_shape(Point(lon, lat), srid=4326)
         property = (
             db.query(SoftStoryProperty)
-            .filter(
-                and_(
-                    geo_func.ST_DWithin(SoftStoryProperty.point, point, 0.000001),
-                    func.lower(SoftStoryProperty.status)
```
nickvisut, line 116: if I understand correctly ...
- before: soft stories that are NOT "work complete, cfc issued"
- after: soft stories

is that right?

backend/api/routers/soft_story_api.py
```diff
@@ -107,21 +108,22 @@ def is_soft_story(
...
+            .filter(geo_func.ST_DWithin(SoftStoryProperty.point, point, 0.000001))
             .first()
         )
 
-        exists = property is not None
-        last_updated = property.update_timestamp if property else None
+        if property:
+            last_updated = property.update_timestamp
+            status_lower = property.status.lower()
+            if status_lower == STATUS_WORK_COMPLETE_LOWERCASE:
```
nickvisut, line 123: and then here we are mapping
- "work complete, cfc issued" to exists=False
- "non-compliant" to exists=True
- everything else to exists=None

backend/api/schemas/soft_story_schemas.py
```diff
@@ -82,7 +82,7 @@ class IsSoftStoryPropertyView(BaseModel):
         last_updated (Optional[datetime]): Timestamp of last update if exists
     """
 
-    exists: bool
+    exists: Optional[bool]
```
nickvisut, line 85: is this to allow for a value of None?

backend/etl/soft_story_properties_data_handler.py
```diff
@@ -77,7 +78,7 @@ def _convert_to_geojson(self, data: list[dict]) -> dict:
         for item in data:
             status = item["status"]
             wkt_point = item.get("point")  # Extract 'point' value (WKT)
-            if status and status.lower() == "work complete, cfc issued":
+            if not status or status.lower() != STATUS_NON_COMPLIANT:
```
nickvisut, line 81: trying to wrap my head around this bit:
- before: skip "work complete, cfc issued"
  - _in other words, include all properties except those w/status "work complete, cfc issued"_
- after: skip empty (or falsy?) status OR any status that is NOT "non-compliant"
  - _does this translate to _including only_ status 'non-compliant'?

backend/api/tests/test_soft_story.py
```diff
@@ -35,8 +35,8 @@ def test_is_soft_story(client, caplog):
     response = client.get(f"api/soft-stories/is-soft-story?lon={lon}&lat={lat}")
 
     assert response.status_code == 200
-    assert not response.json()["exists"]
-    assert response.json()["last_updated"] is None
+    assert response.json()["exists"] is False
```
nickvisut, line 38: mind explaining the change to the test?

nickvisut (commented):
largely makes sense to me and seems to follow what you wrote in the Description. added a few comments/questions just to clarify my understanding of the code.

backend/api/routers/soft_story_api.py
nickvisut: and the reason this change works is that the dataset generated by ETL already excludes them?

backend/api/schemas/soft_story_schemas.py
agennadi: Exactly. False = retroffited, True = non-compliant; everything else defaults to None.

backend/etl/soft_story_properties_data_handler.py
agennadi: Yes, this method generates the static geojson we display on the map (a.k.a soft story points). Before, the "Newly added - Permit Issued" soft stories were displayed as on the map. Now, only valid (skip empty status) and non-compliant (skip everything that is NOT non-compliant) soft stories are saved in the geojson.

backend/api/routers/soft_story_api.py
agennadi: Exactly!

backend/api/routers/soft_story_api.py
agennadi: > before: soft stories that are NOT "work complete, cfc issued"
> after: soft stories

Correct. The original query was "retrieve the soft story if its status != "work complete, cfc issued" (which means, return "non-compliant", "newly added", "permit issued" etc.)
The new query retrieves the soft story regardless of its status. After that, the if-else logic is used to map the status ("work complete, cfc issued" to `exists=False`; "non-compliant" to `exists=True`; everything else to `exists=None`)

> and the reason this change works is that the dataset generated by ETL already excludes them? 

Not really. The change in the ETL workflow only affects the static geojson file. The data in the database is all the same, I just removed filtering by status and changed the mapping logic.

backend/api/tests/test_soft_story.py
agennadi: The original query in the API method was "retrieve the soft story if its status != "work complete, cfc issued" => it didn't return anything if the soft story was retrofitted => the test asserted `response.json()["exists"] = None`.

New query retrieves a soft story regardless of its status and the API returns `{exists: False}` for retrofitted soft stories.

backend/api/routers/soft_story_api.py
nickvisut: I think I see, thanik you!

nickvisut (approved):
thanks for looking into this!

nickvisut (approved):

PR #658, opened 2025-09-24
Add playwright smoke tests for prod site

### LINKED ISSUE #333: Implement Playwright E2E Testing
## Context
It is necessary to create end-to-end tests to make sure key features of the web application keep working as expected. Currently, only partial unit tests are included in the app, which do not provide the necessary coverage. Suggested library for end-to-end testing: Playwright.

## Definition of Done
- Playwright is set up and working in the project
- Test scenarios will cover all major user stories
- Tests are included in the CI/CD pipeline
- This is parent issue, exact test cases will be added in separate sub-issues

## Engineering Details
- Use Playwright for all E2E tests
- Suggested test scenarios :
   - User can view and scroll through the map
   - User can search for an address and see hazard data for the address
   - User can share a report (copy link)
   - User can visit the About page

jessemavi:
-  Related to http://github.com/sfbrigade/datasci-earthquake/issues/333

# Description

Adds the Playwright library and smoke tests for testing critical functionality of the prod SafeHome site. These will be run manually for now but could be changed to run on a schedule(ex. once a day) in the future. Probably will be some trial and error with running the actions workflow to confirm it's working as I'm not an expert but it should at least be close. Tests can be run locally on the prod url with `npx playwright test smoke-tests/`

These tests can also be used in the future for e2e tests in our CI/CD pipeline for pull requests before merging to develop/main. Some updates would need to be made to get the right deployment link and use that as the url to test against. At that point some renaming in the folders/files here and changing from `smoke tests` to `e2e tests` would be better.

Note: Mobile devices are commented out since we are making mobile changes around the visibility of the hazard cards.

See gemini summary below for more info.

## Type of changes
- ( ) Bugfix
- (x) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests

## How to test

## Clean commits


smoke-tests/smoke_tests.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+  await searchBox.click();
+  await searchBox.fill('321 7th Street');
+  await page.getByText('321 7th Street', { exact: true }).click();
+
+  // Expects page to have a heading for the search result address
+  await expect(page.getByText('Report for 321 7th Street')).toBeVisible({ timeout: 10000 });
+  
+  // Expects hazard cards to have pill data for the search result address
+  const cards = page.locator('.chakra-card__root');
+  await expect(cards.first()).toContainText('Compliant');
```
jessemavi: This assertion passes locally but doesn't pass on `develop.dafehome.report` which displays `No Data` for this address. Noticed that the API response doesn't match between local and develop for the `exists` value. I'm assuming that we want to display `Compliant` and not `No Data` so I went with that.

Image
Image

nickvisut: Ahhh I'm realizing now that maybe it's cached on https://develop.safehome.report b/c it seems you have [the commit](https://github.com/sfbrigade/datasci-earthquake/commit/ec4ab40d1b2faf8df4c3f8672db73b4ed8621644) in your branch that fixes this issue. Hmm. Not certain, though.

@agennadi, any ideas on the discrepancy noted above? (not urgent)

nickvisut:
Mind rebasing on latest `develop` when able?

smoke-tests/smoke_tests.spec.ts
jessemavi: Oh actually it was because I needed to restart docker for the backend changes locally. So looks like `No Data` is the right value. Updated the assertion.

Screenshot 2025-09-27 at 1 05 41 AM

agennadi: `No Data` is correct - this address is not in the dataset. I manually purged the cache for soft stories, though (just in case).

nickvisut (approved):
Looks great, @jessemavi!

Let's merge and then see if we can 1) set things up for other environments and/or 2) eventually get it into pipeline to run automatically.

Issue #641, opened 2025-09-16
Update exception message in the APIs


agennadi:
## Context
Including the raw exception message (`str(e)`) in the `HTTPException` detail can leak internal implementation details to the client, which is a security concern. It's better to return a generic error message.

Additionally, including `str(e)` in the log message on line 200 is redundant because `exc_info=True` already captures the full exception details in the traceback.

Refactor the exception block for `is_soft_story()`, `is_in_liquefaction_zone()` `and is_in_tsunami_zone()`.

This is an example for `is_soft_story()`:

```
except Exception:
    logger.error(
        f"Error checking soft story status for coordinates: lon={lon}, lat={lat}",
        exc_info=True,
    )
    raise HTTPException(
        status_code=500,
        detail="An unexpected error occurred while checking soft story status.",
    )
```

blessyparida:
Kindly assign me this issue. I want to fix this.

divya4234:
I want to fix this issue. I request you to kindly assign this to me.

agennadi:
Completed

PR #685, opened 2025-10-15
Update exception for 500 messages for soft story, tsunami zone, liquefaction zone routes

qredwoods:
# Description

_add a short description_
Improves security by removing internal error data from user-facing 500 message for three routes.
_reference to relevant issue_
#641 

## Type of changes
- ( ) Bugfix
- (X ) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests

## How to test

Provoke a 500 error on any of the functions. (unsure how to do this) or review the changes themselves.

## Clean commits
- ( X) I plan to Squash and Merge
- ( X) My commit history is clean¹

### DIFF
--- a/backend/api/routers/liquefaction_api.py
+++ b/backend/api/routers/liquefaction_api.py
@@ -196,10 +196,10 @@ def is_in_liquefaction_zone(
 
     except Exception:
         logger.error(
-            f"Error checking liquefaction zone status for coordinates: lon={lon}, lat={lat}, ",
+            f"Error checking liquefaction zone status for coordinates: lon={lon}, lat={lat}",
             exc_info=True,
         )
         raise HTTPException(
             status_code=500,
-            detail=f"An unexpected error occured while checking liquefaction zone status.",
+            detail=f"An unexpected error occurred while checking liquefaction zone status.",
         )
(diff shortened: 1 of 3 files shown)


qredwoods:
Since the text of each is so similar, I considered creating a function somewhere and importing that into each file, with parameters like lat,lon, and category, and using an f-string variable like f"An unexpected error occurred while checking {category} status." which would be more reusable if other is_something() routes were added later, and also reduce issues like the ones I had above where typos or changes have to be made in three places instead of one. However I don't know the codebase well enough to know where that would go, or if import complexity actually makes sense in this case.

agennadi:
> Since the text of each is so similar, I considered creating a function somewhere and importing that into each file, with parameters like lat,lon, and category, and using an f-string variable like f"An unexpected error occurred while checking {category} status." which would be more reusable if other is_something() routes were added later, and also reduce issues like the ones I had above where typos or changes have to be made in three places instead of one. However I don't know the codebase well enough to know where that would go, or if import complexity actually makes sense in this case.

We are going to add more similar APIs in the future, so keeping exception handling in one place would be helpful! It can be done with a helper function as you suggested - or we could use FastAPI's `@app.exception_handler`. I will research more, create an issue and ping you.

agennadi (approved):
Looks good to me! You asked how to provoke a 500 error - actually, this is something we should write tests for. I created an issue #694 for that

PR #701, opened 2025-11-05
Updates Playwright e2e tests actions workflow to run on successful Vercel preview deployment

jessemavi:
# Description

- Related to https://github.com/sfbrigade/datasci-earthquake/issues/333

This PR does the following:
- Updates actions workflow to run Playwright e2e tests in CI on a successful Vercel preview deployment(when opening a PR or any subsequent commits) or outside of CI when running manually. It sets the env variable to the preview url or an empty string.
- Makes some updates to the playwright config: 
  1) use the env variable(preview url) for the `baseUrl` or fallback to the prod url when the tests are run manually as smoke tests. 
  2) disables Firefox browser from running tests as they are failing in the CI environment. Attempted to fix the errors but wasn't able to so disabling while continuing to investigate. Firefox, by default, is often stricter than Chromium/Webkit in a headless CI environment
- Renames folders/files from `smoke-tests` to` e2e-tests`

You can see this is working from the [check](https://github.com/sfbrigade/datasci-earthquake/actions/runs/19118502815/job/54633462918?pr=701) below. Also the two screenshots below show:
- the vercel preview deployment url from the most recent commit(same url in the most recent vercel bot comment) is being used as the value for `BASE_URL`
- the Playwright trace from running tests on Firefox(commented out now as noted above) shows the vercel preview deployment url is being used and not the prod url

image

image

## Type of changes
- ( ) Bugfix
- (x) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests
- (x) I think tests are unnecessary

## How to test

## Clean commits


.github/workflows/e2e_tests.yml
```diff
@@ -0,0 +1,66 @@
...
+      # Continue even if the tests fail so we can upload the reports/traces
+      continue-on-error: true
+      
+    - uses: actions/upload-artifact@v4
+      # Upload reports regardless of success/failure
+      if: ${{ always() }}
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 1
```
nickvisut, line 56: any downside to keeping these longer?

.github/workflows/e2e_tests.yml
```diff
@@ -0,0 +1,66 @@
...
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v4
+      with:
+        # For deployment_status, checkout the exact commit Vercel deployed.
+        # For manual runs, checkout the current branch reference.
+        ref: ${{ github.event_name == 'deployment_status' && github.event.deployment.sha || github.ref }}
```
nickvisut, line 27: nice!

playwright.config.ts
```diff
@@ -1,10 +1,12 @@
 import { defineConfig, devices } from '@playwright/test';
 
+const baseURL = process.env.BASE_URL || 'https://www.safehome.report';
```
nickvisut, line 3: is https://develop.safehome.report covered?

nickvisut (approved):
lgtm! just left a coupla Qs. is this runnable locally, btw? or is that a future PR?

nickvisut:
@agennadi do you mind taking a look at this PR by @agennadi and associated issue? it's the beginning of e2e tests and includes some smoke tests and happens to do an `npm run build` to set up things for testing.

and cc @jessemavi from your end, note that there is another PR #630 by @agennadi  in review that both does an `npm run build` in CI and makes it more robust across environments that will land soon.

I don't want to block either PR, but I wonder if there may be overlap between the two. If so, does it make sense for them to share code somehow, or is it best that they're isolated from one another? I don't know best practice for GitHub workflows, but I vaguely recall being able to include one .yml in another via my experience with GitLab's equivalent CI configuration. If the former, can we either address in one of these PRs or a future one?

Or perhaps, a third option: The standalone actual execution of `npm run build` is redundant since it's already done in this PR ... 🤷. Please discuss and decide, TIA!

.github/workflows/e2e_tests.yml
```diff
@@ -0,0 +1,66 @@
...
+        
+    - uses: actions/setup-node@v4
+      with:
+        node-version: lts/*
```
nickvisut: FYI the code now has node-version set to 20.x in the `develop` branch in `ci.yml`. It will show up once the branch is updated if it hasn't been already.

On a side note, what is best practice for node-version? being explicit or using `lts/*` here?

It also occurs to me that we use (or will be using) eg `setup-node@v4` action in several places; is there a way to ensure the arguments are the same across the board wherever it's used? @jessemavi cc @agennadi. I think this may be a use case for the code sharing (or at least argument sharing, if that's possible?) that I mentioned in a prior comment. Thoughts?

agennadi: @nickvisut Vercel uses `npm run build` for building preview deployments. Successful deployment triggers e2e tests that also run `npm run build`, so the standalone check in CI might be two times redundant :D but if a deployment fails, e2e tests won't be triggered - while CI will run regardless of deployment status.  

> also occurs to me that we use (or will be using) eg setup-node@v4 action in several places; is there a way to ensure the arguments are the same across the board wherever it's used

We could either use lts or store the node version as an env variable. Btw CI uses `setup-node@v3` and the new workflow uses `setup-node@v4`.

.github/workflows/e2e_tests.yml
jessemavi: Not really assuming we won't run into any storage limits that make us have to pay. I mostly wanted to automatically clear out old artifacts. My assumption was that if a test fails, someone will be downloading the traces and reports immediately to investigate or at least within 24 hrs. If not then they could just re-run the job to see the failure again.

playwright.config.ts
jessemavi: > is this runnable locally, btw?

Yes you can set the `BASE_URL` env locally to `http://localhost:3000` to run them locally.

> is https://develop.safehome.report/ covered?

Not currently but we can run the tests on `https://develop.safehome.report` after code has been merged and deployed successfully if that is something we want to do. It would be easy to add that to this workflow if we already have a custom environment for the develop branch in Vercel. Instead of `Preview` maybe it's something like `Develop` or `Staging`. It looks like [custom environments are a pro feature](https://vercel.com/docs/deployments/environments) though so not sure if we have them.

.github/workflows/e2e_tests.yml
jessemavi: > On a side note, what is best practice for node-version? being explicit or using lts/* here?

I don't have a real strong opinion here and initally went with `lts` so there wouldn't need to be any ongoing updating in this file. But I updated to `20.x` to match what the app uses and the other workflows.

> Btw CI uses setup-node@v3 and the new workflow uses setup-node@v4

Updated this workflow to also use `setup-node@v3`

jessemavi: > happens to do an npm run build to set up things for testing.

I think I'm confused because it doesn't look like the e2e tests workflow I'm adding is doing an `npm run build`. Maybe I'm missing where that is happening. It should just be installing dependencies with `npm ci`.

jessemavi: > I wonder if there may be overlap between the two. If so, does it make sense for them to share code somehow, or is it best that they're isolated from one another?

I don't think there's much overlap between the two workflows and think it's best if they are kept isolated especially since they have different events that trigger them. 1) CI workflow runs immediately and validates code quality and 2) e2e tests workflow runs only after there is a successful preview deployment to test against(or can be run manually) and validates the deployed experience.

nickvisut: OK, that all makes sense. Thanks, @agennadi and @jessemavi, and you're right about us not currently depending on an explicit `npm run build`, so let's keep them isolated.

Can one of you add an Issue to the backlog that helps keep the Node version # (via env variable or shared step/task) consistent across jobs?

nickvisut: @jessemavi This can probably be merged after the latest tweaks. Let's just make sure we don't lose the TODO about the shared Node (and perhaps even action) versions.

agennadi: > Can one of you add an Issue to the backlog that helps keep the Node version # (via env variable or shared step/task) consistent across jobs?

@nickvisut I'll add an issue

PR #717, opened 2025-11-12
Task/680/sync settings env

### LINKED ISSUE #680: Enforce synchronization between `Settings` class and `.env.example`
## Context
Our FastAPI backend uses a centralized `Settings` class (in `backend/api/config.py`) to manage environment variables loaded from the `.env` file. We also maintain a `.env.example` file that serves as a reference for required variables - useful for onboarding new developers and CI setup.

Currently, there’s no safeguard ensuring that new variables added to the `Settings` class are also reflected in `.env.example`. This often leads to inconsistencies where:

- `.env.example` misses required variables;
- CI/CD fails because expected environment variables are undefined;

We need a mechanism to enforce that every field defined in the `Settings` class exists in `.env.example`, even if with a dummy placeholder value.

## Definition of Done
- Implement a validation script (e.g., `scripts/check_env_sync.py`) that:
  - Parses the `Settings` class to extract all defined environment variable names.
  - Reads `.env.example` and compares variable keys.
  - Reports any missing variables.
- Add this check to CI or the pre-commit hook.

## Engineering Details
`Settings.model_fields.keys()` returns a view object containing the names of all the fields defined on a Pydantic settings model.

NewtonLC:
# Description

Added a script `scripts/check_env_sync.py` that:
- Reports Pydantic Validation Error, which occurs when `Settings` is not in sync with the `ENV_FILE` in `backend/api/config.py`
- Parses the `Settings` class in `backend/api/config.py` to extract all defined environment variable names.
- Reads `.env.example` and compares variable keys.
- Reports any variables missing from `.env.example`

Added a pre-commit hook to ensure that this script is run when a developer commits changes to `backend/api/config.py`.

Closes #680 

## Type of changes
- ( ) Bugfix
- (X) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests

## How to test

Edit `backend/api/config.py` and `.env`:
- Add new fields to the `Settings` class in `backend/api/config.py`
- Make sure that field also exists in `.env`, to avoid a pydantic validation error
- Alternatively, comment out fields in the `.env.example` file
  - Make sure to edit `backend/api/config.py` with a comment, print statement, etc. as the hook only runs if this file is changed.

`git add` the changed files, then `git commit` with a test message. 

You should see a message like this in the terminal, at the end of the pre-commit hook messages.
```
Check .env.example matches Settings......................................Failed
- hook id: check-env-sync
- exit code: 1

Missing environment variables in .env.example:
  - TEST_VAR
```

## Clean commits


agennadi (approved):
Looks good, just a few nits: 1) you check that `Settings` exists but don't check if `.env.example` exists. 2) `parse_env_example()` doesn't check lines that aren't comments and don't contain `=`

NewtonLC:
Thanks for the review! I added your suggestions, so the script should be able to check whether `.env.example` exists in the pre-commit hook as well as any malformed lines.

Issue #695, opened 2025-10-22
Add HazardCheckError exception to reduce duplication and improve Sentry logging


agennadi:
## Context
As @qredwoods pointed out, each hazard-zone endpoint (soft story, tsunami, liquefaction) uses repetitive try/except blocks to handle unexpected runtime errors, e.g.:

```
except Exception:
    logger.error(
        f"Error checking soft story status for coordinates: lon={lon}, lat={lat}",
        exc_info=True,
    )
    raise HTTPException(
        status_code=500,
        detail="An unexpected error occurred while checking soft story status.",
    )
```

This approach duplicates code across routers. Moreover, handled HTTPExceptions are not sent to Sentry. Currently, FastAPI's global exception handler (`api/index.py`) only tracks unhandled exceptions of type `Exception` and ignores `HTTPExceptions`.

## Proposed Solution
1. Create a file for custom exceptions in the backend folder. Define a custom exception (e.g. `HazardCheckError`) class that inherits from `Exception`:

```
class HazardCheckError(Exception):
    """Raised when a hazard zone check fails."""
    def __init__(self, zone: str, lon: float, lat: float, original_exception: Exception = None):
        self.zone = zone
        self.lon = lon
        self.lat = lat
        self.original_exception = original_exception
```

2. Raise `HazardCheckError` instead of `HTTPException` in `is_in_liquefaction_zone()`, `is_in_tsunami_zone(), is_soft_story()`.

3. In `api/index.py` create a new exception handler **before** `global_exception_handler()`. The new exception handler should: 1)  have the `@app.exception_handler(HazardCheckError)` annotation, 2) print the logs 3) send the error to the Sentry server. Use `global_exception_handler()` as a reference. 

4. Create `backend/api/tests/unit/test_hazard_check_error_handler.py` and add a unit test that mocks Sentry and simulates a `HazardCheckError` exception to check that the exception handler behaves correctly.

5. Create `backend/api/tests/test_sentry_capture.py `and add an integration test that mocks Sentry and a database session and simulates a generic exception (e.g. SQLAlchemyError) during a hazard API request.

PR #728, opened 2025-11-15
enhancement/715/parallelize-etl-gha

santoshgdev:
# Description

To palatalize the ETL steps, i've moved before and after it into their own steps and made the actual etl step a matrix strategy.

## Type of changes
- ( ) Bugfix
- ( ) Chore
- (x) New Feature

## Testing
- ( ) I added automated tests

## How to test

## Clean commits
- (x ) I plan to Squash and Merge

### DIFF
--- a/.github/workflows/etl_to_neon.yml
+++ b/.github/workflows/etl_to_neon.yml
@@ -7,12 +7,8 @@ on:
   workflow_dispatch:  # Allows manual triggering of the workflow
 
 jobs:
-  run-etl:
-    runs-on: ubuntu-latest  
-    outputs:
-      changes: ${{ steps.detect_changes.outputs.changes }}    
-      etl_commit: ${{ steps.detect_changes.outputs.commit_sha }}
-
+  pre-etl:
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


qredwoods (commented):
Thanks Santosh! Defer to Anna on approval, but the condensing lines into a structure we iterate over like in line 69 LGTM and is a good style/prep for expansion, expect pattern will recur many times across project I think since I'm guessing there was a history of adding one data set at a time. 


Tangent - (and sorry if my terms are blurry here) - Maybe we can generalize that to a style guideline to encourage it specifically around when handling that expected to grow trio - there was also discussion in the case of a prior issue about trying to abstract at a higher level of things where liquefaction, soft_story, and tsunami were swapped in. In the future I think @agennadi we could look at if there's a design possible that minimizes the need to add an nth thing in every place in the code when an nth data layer is added by maybe making it a single reference that automatically goes through all the layers/sets. Maybe the principle is "when something touches all of the data layers (ie tsunami, soft-story AND liquefaction), take advantage of opportunities to limit repetition of code" as Santosh has done here.

qredwoods:
I saw Anna's thumbs up on my comment above -- if that means @agennadi  you support PR approval I can go ahead and approve this or you also can directly.

qredwoods:
(Also to follow up on my point above about standardizing that we always try to combine the layers -- in tonight's meeting, I was successfully convinced it doesn't make sense to do that. Anna pointed out that in other parts of the code the treatment of each dataset is necessarily very distinct due to the differences between datasets -- for example liquefaction and tsunami zones being mapped polygon based, whereas soft-story is address-based data. Therefore when there's code duplication like here or moreso #695, sure, take opportunities to consolidate, but let's not design away the flexibility in pursuit of arbitrary unity, especially as each new layer will bring additional unique traits that require special-case handling. )

agennadi (changes requested):
I ran the workflow from a test branch (`715/test`) and the jobs are failing (check out the logs here: https://github.com/sfbrigade/datasci-earthquake/actions/runs/19606231119). 

This happens because each github job has an isolated environment. The `pre-etl` job installs python and the dependencies but `run-etl` knows nothing about them and the job fails with `uv: command not found error`. 

`post-etl` fails with `fatal: not a git repository (or any of the parent directories): .git` for the same reason - there is no repository checkout in this job. 

Thus, each job would require its own setup and/or checkout. Otherwise, we could give up the matrices and go with bash parallel execution using `&` - what do you think?

PR #766, opened 2026-01-01
Update README.md

oscarsyu:
Not a major push.  Just adding in links to previous analysis done to validate datasets used

### DIFF
--- a/datasets-eda/README.md
+++ b/datasets-eda/README.md
@@ -2,5 +2,7 @@ This folder contains notebooks for exploratory data analyses
 
 - fema_nri: [FEMA National Risk Index](https://hazards.fema.gov/nri/)
 - usgs_faults: [USGS Quarternary Faults database](https://www.usgs.gov/programs/earthquake-hazards/faults)
-
+- seismic hazard zones: [https://colab.research.google.com/drive/1JrXsDYjmp0iLEHOfSapWvT1TNEYLyyHI?usp=sharing](https://colab.research.google.com/drive/1k6OP9QVcVtxVXDbiz4pXXD2xIwq8SDY6?usp=sharing)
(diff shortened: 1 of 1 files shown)


qredwoods:
These all look good to include. Optionally consider including:
1."Requires project Google Drive access" or something similar. 
2. 'You may need to select in top right "Open with Google Collaboration"

When I clicked these links I first got the "You need to request access" and then I got a bizarre looking Drive view with just one small jpeg and 30 comments. Once I switched to the right drive account and clicked on Open with Google Collaboration everything looked good.

Only fix need for approval from my opinion: note the google gemini responses about the mix up of the link text versus the actual link for the first link. 

If you just want to commit that suggestion or change it yourself either way is fine (I'm neutral on having full text of link as the the link text vs a descriptive phrase). TY!

qredwoods:
Also I noticed that the original links in there, the fed gov ones, now give 403 Forbidden errors... perhaps those are no longer public or have been changed?

oscarsyu:
> Also I noticed that the original links in there, the fed gov ones, now give 403 Forbidden errors... perhaps those are no longer public or have been changed?

Thanks @qredwoods for taking a look!  Added a description about getting gdrive access. The fed gov links still works for me.  could you confirm that's the case and if it's fine, help merge that in?

qredwoods (approved):
improves datasets-eda/README.md by directing readers to additional notebooks the project team has explored. Is well formatted and all links are correct.

qredwoods:
Unfortunately, as addressed in #782, we are getting blocked by docker build timeouts - way overkill for .md only changes. I will open a new issue and pr to unblock .md only changes from requiring CI pipeline, as they do not change behavior anyway and nothing in the pipeline affects them.

qredwoods:
Ok @NewtonLC 's fix on the larger pipeline made this one able to pass, TY! Will still simplify the .md approach for the future (#793 )

Issue #782, opened 2026-01-23
Consolidate and Clean Up Docker & Initialization Logic


NewtonLC:
## Context
The ETL scripts are currently executed on every push and PR via CI. We suspect that this was originally set up to ensure test data stayed up-to-date, particularly when some datasets (e.g. Soft Stories) saw daily updates.

However, in a discussion with @bees-dont-have-knees, @qredwoods, and @CarlyAThomas, we agreed that our approach appears to be unnecessarily expensive:

- Slow pipelines: ETL work adds on average 20-40 extra seconds of processing time to every PR and push.
- Wasted resources: CI time, compute, and network usage are consumed.
- Unneeded ETL runs: The backend ETL scripts are re-run on **every change**, including frontend changes and documentation.
- Low dataset churn: 
  - Tsunamis dataset last updated in 2022
  - Liquefactions dataset last updated in 2024
  - Soft Stories last updated July 2025, and no longer appears to update regularly

Additionally, SafeHome is considering incorporating more datasets in the future, which will escalate these issues.

We are considering removing ETL execution from CI entirely and instead relying on static inserts or seeded data for tests, while decoupling ETL runs from the PR pipeline.

## How to Reproduce
1. Open any pull request or push a commit to a branch with CI enabled. (Even just adding "hello" somewhere to the README.md)
2. Observe that the ETL scripts are executed as part of the CI pipeline.
3. Note the additional time added to the pipeline due to ETL execution, even when changes do not affect data ingestion or schemas. (This can be viewed in the **Docker Logs** tab in the `docker-build-test` job. The ETLs to look out for are TsunamiDataHandler, LiquefactionDataHandler, and SoftStoryPropertiesDataHandler)
4. Repeat across multiple PRs to see consistent CI slowdowns and redundant ETL runs.

## Definition of Done
- ETL execution is removed from the default CI pipeline for PRs and standard pushes.
- Tests are updated to rely on explicit inserts, seeded data, or fixtures instead of requiring a full ETL run.
- A clear mechanism exists to run ETL intentionally and separately (e.g. manual trigger, scheduled job, or dedicated workflow).
- CI pipelines are measurably faster and limited to linting, builds, and tests where appropriate.
- Docker and initialization logic are updated as needed to reflect the new separation of concerns.
- Documentation is added explaining:
  - Why ETL was decoupled from CI
  - How test data is now provisioned
  - How and when ETL should be run going forward
  - Any implications for CI, Docker, and local development
[...]

PR #794, opened 2026-01-26
[Temporary] Increased docker compose up timeout duration

NewtonLC:
# Description

Increased --wait-timeout parameter in Start Services job to 5 minutes
Increased the number of retries per healthcheck and shortened interval between healthchecks.

This is a **TEMPORARY** band-aid fix to this issue, where the "docker compose up" command fails to return a healthy response before the 2-minute timeout duration. Investigating and fixing docker initialization logic has been noted as issue #782. However, in the mean-time, we need a temporary fix to allow CI to return a healthy response.

## Clean commits
- (X) My commit history is clean¹

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,7 @@ jobs:
         run: docker compose build
 
       - name: Start Services
-        run: docker compose up -d --wait --wait-timeout 120
+        run: docker compose up -d --wait --wait-timeout 300
 
       - name: Docker Logs 
         if: always()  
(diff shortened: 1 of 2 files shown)


qredwoods (approved):
LGTM.
Changes:

- increases --wait-timeout to 5 minutes

- critically reduces the health_check interval from 500s (8mins 20s) to 10s. Since backend indeed depends on db being healthy, this closes a big time problem if the db isn't healthy on the first check. (For local dev, afaik the health checks run the entire life of the container, but only as health metadata, so the every 10s curl calls should not pose issues and to an extent improve metadata clarity by narrowing the interval of checks)

- increases the number of retries, by pairing this with shorter intervals it gives a lot of healthcheck retries over 1-2 minutes, giving fast feedback in the event of actual failure

PR #819, opened 2026-02-15
Create etl_tests.yml

bees-dont-have-knees:
# Description

add new workflow, separate from the CI workflow, to run integ tests for ETL.  will eventually set up rule to have this run on merges to develop and main (or maybe just main...TBD). it's a work in progress so for now, it will be setup for manual triggering for testing purposes.

## Type of changes
- ( ) Bugfix
- ( ) Chore
- (x) New Feature

### DIFF
--- a/.github/workflows/etl_tests.yml
+++ b/.github/workflows/etl_tests.yml
@@ -41,7 +41,6 @@ jobs:
         id: check_fork
         if: github.event_name == 'pull_request'
         run: |
-          run: |
             if [[ "${{ github.event.pull_request.head.repo.full_name }}" != "${{ github.repository }}" ]]; then
               echo "is_fork=true" >> $GITHUB_ENV
             fi
--- a/.github/workflows/etl_tests.yml
+++ b/.github/workflows/etl_tests.yml
@@ -1,0 +1,106 @@
+name: ETL tests
+
+on: workflow_dispatch
+  
+jobs:
+  
+  black_lint_and_mypy_type_check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/etl_tests.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: docker compose up -d --wait --wait-timeout 30 db
+
+      - name: Docker DB Logs 
+        run: docker logs my_postgis_db 
+
+      - name: Run Tests
+        run: echo 'Coming soon'
+
+      - name: Clean Up
+        run: docker compose down db
```
NewtonLC, line 105: Should just be `docker compose down`

.github/workflows/etl_tests.yml
```diff
@@ -0,0 +1,106 @@
...
+        uses: actions/checkout@v4
+
+      - name: Set default fork status
+        run: echo "is_fork=false" >> $GITHUB_ENV        
+
+      - name: Check if pull request from fork
+        id: check_fork
+        if: github.event_name == 'pull_request'
+        run: |
+          run: |
```
NewtonLC: We can't have duplicate `run: |` lines, and this error also occurs in ci.yml so that needs to be fixed as well.

NewtonLC (changes requested):

.github/workflows/etl_tests.yml
bees-dont-have-knees: i made it explicit since the only service that's req for ETL is "db".  compose up db => compose down db

https://docs.docker.com/reference/cli/docker/compose/down/

NewtonLC (approved):
lgtm!

Issue #829, opened 2026-02-21
Upgrade to Next 16


nickvisut:
We are upgrading to Next 16 to keep up with the latest features as well as reduce vulnerabilities across associated packages.

PR #854, opened 2026-02-28
remove next-env.d.ts entirely and fix reference to it in .gitignore file

### LINKED ISSUE #781: Remove `next-env.d.ts` from repo
As per docs (https://nextjs.org/docs/app/api-reference/config/typescript#next-envdts) and as a request from @qredwoods, we should remove `next-env.d.ts` from repo. It seems to be regenerated differently on e.g. `dev` or `build`, polluting local commits (and possibly PRs).

Add `next-env.d.ts` to the git.ignore to prevent errors from arising in the future.

nickvisut:
# Description

#827 was a swing and a miss. turns out there isn't a way to get .gitignore to ignore a file already in the repo. it's either committed or it's not, no in between as has been suggested and LLMs seem to affirm (yet another reason to not rely on AI summaries without checking sources thoroughly). however, it's my mistake for not testing properly!

this PR thus both removes AND deletes next-env.d.ts which now works since the CI permissions issue was fixed as part of #829.

Closes #781 

## Type of changes
- ( ) Bugfix
- (x) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests
- (x) I think tests are unnecessary

## How to test

create PR with change to `next-env.d.ts` and it shouldn't show up in `git status`

## Clean commits
- (x) My commit history is clean¹

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # Next
 .next
 out
-next-env.d.ts # TODO: look into removing this file from source control entirely
+next-env.d.ts
 
 # Python
 __pycache__/
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,8 +22,8 @@ RUN mkdir -p /app/.next && chown -R node:node /app/.next
 
 USER node
 
-# Copy package*.json, next-env.d.ts, nstall dependencies, as node user
-COPY --chown=node:node ./package*.json next-env.d.ts ./
+# Copy package*.json and install dependencies, as node user
+COPY --chown=node:node ./package*.json ./
 RUN npm install
 
 # Copy the rest of the application, ensuring the ownership is set to node user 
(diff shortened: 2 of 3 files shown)


nickvisut:
> ## Code Review
> 
> This pull request correctly removes the `next-env.d.ts` file and updates the `.gitignore` file. However, a critical step was missed. The `tsconfig.json` file still contains a reference to the now-deleted `next-env.d.ts` in its `include` array. This will cause the TypeScript build to fail. To fix this, you need to remove `"next-env.d.ts"` from the `include` array in `tsconfig.json`.

attempting to follow Next docs. may be okay to leave it in `tsconfig.json`. if not, follow-up PR with, for example, `next typegen` may help.

ZaimaGull (approved):

PR #880, opened 2026-03-26
Allow CI to trigger on branches prefixed with '/ci'

NewtonLC:
# Description

Now, CI will trigger automatically on-push if the branch name is prefixed with `/ci`. Previously, only protected branches(`main`, `develop`) had CI trigger on push. Developers would have to create a PR in order to test the CI, which is less convenient.

This change allows developers to test CI **without needing to create a PR.** 

## Type of changes
- ( ) Bugfix
- ( ) Chore
- (X) New Feature

## Testing
- ( ) I added automated tests

## Clean commits
- (X) My commit history is clean¹

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - develop
       - main
+      - 'ci/**'
   pull_request:
     types: [opened, synchronize, reopened]
 


.github/workflows/ci.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - develop
       - main
+      - 'ci/**'
```
nickvisut, line 8: lgtm. found my branch, but approach was way more complex and don't recall why.

nickvisut (approved):

PR #910, opened 2026-04-19
vercel preview builds on demand

bees-dont-have-knees:
# Description
https://github.com/sfbrigade/datasci-earthquake/issues/911

currently, every commit kicks off a Preview build, whether you want it or not.  however, depending on the task, you might not care about Preview builds (for example, if your task revolves around github workflows or if you're working on ETL).  

this new change adds the ability to toggle vercel preview builds via commit message. the default is NOT to deploy to preview.  if you wish to deploy to preview, you need to add "[deploy-vercel-preview]" to your commit message.  

ultimately, this will reduce the number of unnecessary preview builds.  and, if at some point, vercel decides to start billing for preview builds, this PR could help keep our costs down.

## Type of changes
- ( ) Bugfix
- ( ) Chore
- (X) New Feature


NewtonLC:
Hello, thank you for working on this change! After running these changes by the frontend team, I received the following feedback:
- Preview deployments shouldn't be opt-in, and requiring a specific tag in the commit message can pollute commit history
- Preview deployments should be required for PR's, so it isn't possible to bypass our e2e safeguards.

I'm going to look into an alternate approach that makes preview deployments PR-only using GitHub Actions.

Issue #911, opened 2026-04-19
vercel preview builds on demand


bees-dont-have-knees:
currently, every commit kicks off a Preview build, whether you want it or not. depending on the task, you might not care about Preview builds (for example, if your task revolves around github workflows or if you're working on ETL). this new change will reduce the number of unnecessary preview builds.

NewtonLC:
PR #910 got feedback from the frontend team:
https://github.com/sfbrigade/datasci-earthquake/pull/910#issuecomment-4284278609

We'll try tackling this approach from another angle. I'm going to try making changes to Vercel's Ignored Build Command settings and potentially our GitHub Actions.

PR #912, opened 2026-04-20
Test commit for Vercel Preview Deployments

NewtonLC:
# Description

NOTE: This PR has been updated. Review past versions by clicking the `edited` dropdown at the top.

Related Issue: #911 

Currently, each time a branch is opened or a commit is made, Vercel automatically makes a preview deployment. This is an excessive use of preview deployments. In most cases, SafeHome developers use preview deployments for manual and e2e testing.

Solution: We want to only deploy to Vercel on PRs and pushes to `develop` and `main`. We will stop Vercel's automatic deployments and instead run them through a GitHub workflow.

**Part 1: GitHub Workflow**
This PR adds a new workflow that runs
- when a PR is opened, re-opened, or pushes a commit.
- when a commit is pushed to the `develop` or `main` branch

**Part 2: Disable Vercel's automatic deployment**
After this PR is merged, I will change the settings in Vercel: `Ignore Build Step` needs to change from "Automatic" to "Don't build anything".
This command prevents Vercel from automatically deploying on each commit. 

## Result:
- Production deployments are NOT affected
- Vercel only makes Preview deployments when:
  - a PR is opened
  - a commit is pushed to a PR

## Type of changes
- ( ) Bugfix
- (X) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests

## How to test

Vercel will not deploy from a branch unless it has a PR open. Upon opening a branch, verify that this is the case:

### On Vercel's dashboard:
Screenshot 2026-04-20 143647
Screenshot 2026-04-20 143721
Screenshot 2026-04-20 143751

### Additionally, you can open a new branch sourced from `chore/newtonlc_skip_preview_deploy`. Upon creating a PR from your test branch to `chore/newtonlc_skip_preview_deploy`, verify that a Preview deployment is made from the HEAD.
Screenshot 2026-04-20 162316


.github/workflows/vercel-pr-open-preview.yml
```diff
@@ -0,0 +1,39 @@
...
+  deploy-preview:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - uses: actions/setup-node@v4
```
nickvisut, line 17: can you replace this with `setup-node-using-nvmrc`? (see other usages); it attempts to centralize e.g. which node version is being used.

feel free to bump to v4 or the the latest `setup-node`.

.github/workflows/vercel-pr-open-preview.yml
```diff
@@ -0,0 +1,39 @@
+name: Vercel Preview On PR Open
+
+on:
+  pull_request:
+    types: [opened, reopened]
```
nickvisut, line 5: should `synchronize` also be in here? it appears that `[opened. synchronized, reopened]` [is the default for PRs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request ... see first bullet point under "Note:") and `synchronize` happens [when pushes to the source branch are done](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=synchronize#pull_request) (double check this, ofc).

nickvisut (changes requested):
Looks mostly good to me. Added a couple of comments.

btw, how does Vercel's Ignore Build Step work in conjunction with the Github Action? If the Github Action instructs Vercel to build and deploy only for PRs, then what is Ignore Build Step preventing? Is having Ignore Build Step set to always ignore separate from the Github Action or is there some overlap?

.github/workflows/vercel-pr-open-preview.yml
nickvisut: could also use this opportunity to update the other `actions` versions too, perhaps

nickvisut:
> Looks mostly good to me. Added a couple of comments.
> 
> btw, how does Vercel's Ignore Build Step work in conjunction with the Github Action? If the Github Action instructs Vercel to build and deploy only for PRs, then what is Ignore Build Step preventing? Is having Ignore Build Step set to always ignore separate from the Github Action or is there some overlap?

ah, I just remembered:
1) deployment to `develop.safehome.report`
2) deployment to prod

so perhaps this Ignore Build Step might affect those.

in the description for the PR, it's stated that "production deployments are not affected". mind explaining how that works?

NewtonLC:
Let me answer some clarifying questions!

> How does Vercel's Ignore Build Step work in conjunction with the Github Action? If the Github Action instructs Vercel to build and deploy only for PRs, then what is Ignore Build Step preventing? Is having Ignore Build Step set to always ignore separate from the Github Action or is there some overlap?

The way Vercel's Ignore Build Step works is, every time a commit is pushed to any branch, it will run that command to determine whether or not it should deploy. It's currently preventing pushes to branches that do not have a PR made.

> Should synchronize also be in the .yml?

The Ignore Build Step command of `if [ -z "$VERCEL_GIT_PULL_REQUEST_ID" ]; then echo "Skip: not a PR"; exit 0; else echo "Deploying PR #$VERCEL_GIT_PULL_REQUEST_ID"; exit 1; fi` is effectively handling the same thing.

However, you have a point, it's confusing for me to put half of the logic in a Vercel command and the other half in GitHub Actions. I will update this PR to move the logic entirely into GitHub Actions, since Vercel alone cannot do what we need!

> How are production deployments not affected?

You're right, this is a mistake! My reasoning was that since both `develop` and `main` are protected and require pull requests, they would be covered in this change. But after double-checking, I do need to include any pushes to `develop` and `main`.

NewtonLC:
There's also a minor tradeoff to deploying Vercel from GitHub Actions: 

Screenshot 2026-04-22 183123

**The top deployment is the GitHub Actions-Initiated deployment, and the bottom two are Vercel's automatic deployments.**

On Hover:
Screenshot 2026-04-22 183145
image

When Vercel deploys automatically, it uses webhooks from GitHub that includes author metadata. But there isn't a way to invoke this through the Vercel CLI in GitHub Actions.

This does not affect development because we still can see who authored the commit, and because we can access the pull request directly.

.github/workflows/vercel-pr-open-preview.yml
NewtonLC: Done!

nickvisut (approved):
lgtm!

Issue #492, opened 2025-07-11
Implement composite API endpoint for multi-hazard lookup


agennadi:
## Context
Currently, the frontend must make separate API calls to retrieve hazard data (soft story, liquefaction, tsunami) for a given address. This increases latency and complexity. A composite API endpoint will allow the frontend to request all supported hazard types in a single call, returning a unified response. 

## Definition of Done
- A new API endpoint (e.g., `/api/hazards/lookup`) is implemented in the `backend/api/routers` folder (create a new python file).
- The endpoint accepts coordinates (lon, lat).
- The endpoint queries all supported hazard types (soft story, liquefaction, tsunami) for the given location.
- A pydantic schema is created for the composite response (e.g. `CompositeHazardResponse` containing `HazardStatus` schema for each hazard type).
- The endpoint returns a single JSON response containing the status (`exists`) and `last_updated` for each hazard type.
- The router is included in `api/index.py`. 
- Unit tests are provided. 
- Swagger documentation for the new endpoint is provided.

## Engineering Details
- Use existing API functions in `backend/api/routers` as a reference.
- Proposed response schema:
```
class HazardStatus(BaseModel):
    exists: bool
    last_updated: Optional[datetime] = None

class CompositeHazardResponse(BaseModel):
    soft_story: HazardStatus
    liquefaction: HazardStatus
    tsunami: HazardStatus
```
- API tests use mock data from the `test_db` docker container. The test data is populated by the backend/database/init.sql script. Feel free to adjust/add more test data.

PR #968, opened 2026-07-09
Add composite hazard lookup API endpoint Issue #492

Neelansh-Khare:
# Description
Added a composite hazard lookup API endpoint (GET /api/hazards/lookup) that returns soft story,
liquefaction, and tsunami status for a given lon/lat in a single response, so the frontend no
longer needs to make three separate API calls per address.

# Type of changes
( ) Bugfix
( ) Chore
(X) New Feature

# Testing
(X) I added automated tests
( ) I think tests are unnecessary

How to test
Start the stack: docker compose up -d --build
Run the unit tests: docker compose exec backend pytest backend/api/tests/test_hazard_lookup.py -v
  Verify all 6 tests pass (all-present, partial, none-present, ping, missing-params,
  database-error).
Manually hit the endpoint:
  curl "http://localhost:8000/api/hazards/lookup?lon=-122.41211&lat=37.80541"
  Confirm the response contains soft_story, liquefaction, and tsunami keys, each with
  exists and last_updated.
Check ping mode: curl "http://localhost:8000/api/hazards/lookup?ping=true" returns
  exists=false for all three hazards with no DB hit.
Check Swagger docs: open http://localhost:8000/docs and confirm /api/hazards/lookup
  is listed under the "hazards" tag with CompositeHazardResponse as the response schema.

# Clean commits
( ) I plan to Squash and Merge
(X) My commit history is clean¹

## Summary by CodeRabbit

* **New Features**
  * Added a hazard lookup endpoint that checks soft-story, liquefaction, and tsunami risks for a geographic location in one request.
  * Added coordinate validation and a ping mode for connectivity checks.
  * Responses indicate whether each hazard is present, when it was last updated, and whether a check failed.

* **Tests**
  * Added coverage for complete, partial, unavailable, failed, and connectivity-check results.


backend/api/routers/hazard_lookup_api.py
```diff
@@ -0,0 +1,138 @@
...
+    if lon is None or lat is None:
+        logger.warning("Missing coordinates in non-ping request")
+        raise HTTPException(
+            status_code=400,
+            detail="Both 'lon' and 'lat' must be provided unless ping=true",
+        )
+
+    logger.info(f"Checking composite hazards for coordinates: lon={lon}, lat={lat}")
+
+    try:
```
NewtonLC, line 130: The three check calls should get their own try/catch block, otherwise if one of them fails, it will fail for all of them. I think this specific exception should only occur if all of the lookups fail, and if only one of them fails then it returns a partial result?

This may be a larger fix though, since it would also need a change to the hazard_lookup_schema so it can give a "check failed" response, plus new tests.

NewtonLC (changes requested):
Hi Neelansh, thanks for working on this issue! A few notes:

Is the landslide implementation md file needed? It looks like it's meant for a different issue.

Neelansh-Khare:
@NewtonLC should be good to review again!

Issue #977, opened 2026-07-24
Turn off chakra's globalCss layers or look into moving some of the CSS into a separate layer if possible


nickvisut:
We use !important in `theme.ts` globalCss because it's using layers. we can get rid of !important if we turn off layers or look into putting some of the CSS into a different layer than the base one.

> should be fine for Chakra UI v3. we may want to turn off layers later, though, or put some of this on a different layer

_Originally posted by @nickvisut in https://github.com/sfbrigade/datasci-earthquake/pull/976#discussion_r3643207277_

kgujral2612:
Pasting notes from my discussion with @nickvisut on slack: 

The options are:

1. Disable Chakra layering: we have minimal style overrides, and introducing layers for a few styles is not necessary imo. I like this option better.
2. Add a top layer, putting the override styles in that. We can get rid of !important  for the Chakra styles. Layered styles would still have a lower specificity than the unlayered mapbox-gl styles.

I am going to create a PR to implement option 1.

kgujral2612:
Other notes that are the definition of done for this issue: 
- `!important` should not be used to override the styles
- The styles should be applied as before this change

kgujral2612:
PR: https://github.com/sfbrigade/datasci-earthquake/pull/985

PR #985, opened 2026-08-06
Disable chakra layers

kgujral2612:
# Description

Disables Chakra layers to get rid of the `!important` keyword while still preserving the styles. I've added notes inside the issue around my research on the options.

Issue: https://github.com/sfbrigade/datasci-earthquake/issues/977

## Type of changes
- ( ) Bugfix
- (X) Chore
- ( ) New Feature

## Testing
- ( ) I added automated tests
- (X) Since these are UI changes, I tested by inspecting DOM elements, finding the computed styles, and comparing that against what is expected per the code. I could not see any difference with and without my changes in the UI, but I am open to other suggestions on testing this. 

## How to test

Go to the home page > Ensure that the zoom in and out buttons appear the same size and that the search dropdown is docked

## Clean commits
- ( X) I plan to Squash and Merge (if I add another revision after a review)
- ( X) My commit history is clean¹

### DIFF
--- a/styles/theme.ts
+++ b/styles/theme.ts
@@ -279,7 +279,7 @@ const globalCss: SystemConfig["globalCss"] = {
     fontFamily: "body", // This applies Inter (the "body" token) globally
   },
   ".chakra-drawer__positioner": {
-    zIndex: "docked !important",
+    zIndex: "docked",
   },
   ".marker": {
     width: "mapMarkerWidth",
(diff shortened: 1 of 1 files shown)


kgujral2612:
I'm getting more acquainted with the project; https://develop.safehome.report/ is a good baseline to compare the UI against. I'll do a more comprehensive test and post notes/push commits.

PR #1021, opened 2026-09-01
Add lightweight feature issue form

nickvisut:
## Summary
- add a structured Feature issue form
- require only the problem at intake
- keep outcome, acceptance criteria, scope, and background optional for refinement
- set the issue type to Feature and apply `needs triage`

## Why
The existing feature request template asks submitters to propose a solution up front. This form keeps intake focused on the problem while allowing the feature to mature during triage and refinement.

## Validation
- form added at `.github/ISSUE_TEMPLATE/feature.yml`
- based on the current protected `develop` head
- live rendering can be verified after merge because GitHub serves issue forms from the default branch

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature.yml
+++ b/.github/ISSUE_TEMPLATE/feature.yml
@@ -1,0 +1,73 @@
+name: Feature
+description: Propose a product feature or enhancement
+title: ""
+type: Feature
+labels: ["needs triage"]
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Start with what you know. Only the problem is required.
+        Other sections can be filled in during refinement.
+
+  - type: textarea
+    id: problem
+    attributes:
+      label: Problem
+      description: >-
+        Who has the problem, what are they trying to do, and what is missing today?
+      placeholder: |
+        When I'm looking at my address's results and see a hazard term I don't
+        recognize, I want to understand what it means so I can judge whether
+        it changes my decision.
+
+        Nothing on the page defines these terms, and the official sources are
+        written for engineers.
+    validations:
+      required: true
+
+  - type: textarea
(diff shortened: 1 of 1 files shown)


nutanp (commented):
Looks good

.github/ISSUE_TEMPLATE/feature.yml
```diff
@@ -0,0 +1,73 @@
+name: Feature
+description: Propose a product feature or enhancement
+title: ""
+type: Feature
+labels: ["needs triage"]
+
+body:
+  - type: markdown
```
radhika-designs, line 8: Might not be required?

.github/ISSUE_TEMPLATE/feature.yml
```diff
@@ -0,0 +1,73 @@
...
+
+        Nothing on the page defines these terms, and the official sources are
+        written for engineers.
+    validations:
+      required: true
+
+  - type: textarea
+    id: desired-outcome
+    attributes:
+      label: Desired outcome
```
radhika-designs, line 33: Maybe add a line that this is not equivalent of design, Eng spec?

radhika-designs (changes requested):
Can you review comments?

.github/ISSUE_TEMPLATE/feature.yml
nickvisut: I agree, can we remove it in a follow-up?

.github/ISSUE_TEMPLATE/feature.yml
nickvisut: Sounds like a good idea. In follow-up as well since we want to land this right away.

radhika-designs (approved):

nutanp (approved):

Issue #1015, opened 2026-08-27
CityCamp event prep


leela-solomon:
These are items we ideally address ahead of the CityCamp event which takes place around 9/19.

nickvisut:
@leela-solomon I added a quick snippet in the description above, but I may have gotten the date wrong.

Issue #990, opened 2026-08-06
Search result marker is hidden by the drawer at some screen sizes


radhika-designs:
**Describe the bug**
Upon searching, the resulting destination-marker is centered on the map. However, it is covered/hidden by the drawer at some screen sizes (see attached video under Screenshots).

**To Reproduce**
1. Go to 'https://develop.safehome.report/'
2. Use search bar and search acceptable address by typing any 3 numbers and pick one address from the options from the dropdown
3. Bingo! -See error in video
4. If on your mobile screen the destination/search result **marker** appears to not automatically appear at the center of the screen then that is the bug (however, if you pan, you'll see that the marker is actually underneath the drawer).

**Expected behavior**
A search result marker on the map should be visible at the center of the map after hitting the enter/searching result.

**Screenshots**
https://github.com/user-attachments/assets/1f329eda-2c53-4e09-bbab-52411ce6b7f3

**Smartphone (please complete the following information):**
- Device: iPhone14ProMax
- Browser: in-app browser, but also happens in mobile Safari

**Additional context**
- map marker is present, but offscreen (underneath the drawer)
- params in URL look fine, unlike with other issue #987
- the hazard data loads, also unlike #987

nickvisut:
contributed by @nutanp:
https://github.com/user-attachments/assets/a85607a3-0ccb-44f2-b161-6079ff4de42f

iPhone 14 Pro Max, just like original issue

Safari

radhika-designs:
https://github.com/user-attachments/assets/f2b1efb3-58fc-45d5-9245-2c2350595e66

radhika-designs:
https://github.com/user-attachments/assets/7c4a633f-c583-44a0-81fb-4243ea9b26a6

PR #1026, opened 2026-09-02
Center map marker within the visible area above the mobile drawer (#990)

danielATiriai:
# Description

Follow-up to #1025 for #990 (tracked by #1015). With the canvas fix in #1025, the search marker's tip sits exactly on the bottom drawer's top edge on phones, because the map centers on the whole container while the drawer covers its bottom half. This change centers the marker in the visible (undrawered) area on mobile and re-centers when the drawer is closed or reopened.

**How.** mapbox-gl treats `center` as the center of the *padded* viewport, so the map now gets `padding.bottom` equal to the drawer's share of the container while the drawer is open on mobile:

- `drawer.tsx` becomes controlled (`open` / `onOpenChange`) and exports `MOBILE_DRAWER_HEIGHT_RATIO`, which is paired with its `h="1/2"` token. The two icon-only buttons gain aria-labels.
- `address-mapper.tsx` owns the drawer state, derives the ratio from `useBreakpointValue({ base: true, md: false })` and drawer state, and passes it to `Map`.
- `map.tsx` applies the padding on creation, on every pan (`easeTo` with `padding`), when the ratio changes (eased, or instant before the style has loaded), and after each `ResizeObserver` resize since padding is in pixels.

Desktop is unaffected: the drawer is on the left there, so the ratio is 0.

Stacked on #1025; the diff includes that commit until it merges, after which this branch will be rebased.

## Type of changes
- (x) Bugfix
- ( ) Chore
- ( ) New Feature

## Testing
- (x) I added automated tests

- `address-mapper.test.tsx`: the Map receives ratio 0.5 on mobile with the drawer open, 0 after closing it, 0.5 after reopening, and 0 on desktop.
- `map.test.tsx`: `setPadding` on creation, padding re-applied after an observer resize, `easeTo` with padding on pan, padding 0 when the ratio is 0.

## How to test

1. Open the preview at a phone viewport (e.g. iPhone 14 Pro Max) with `/?address=123+Main+St&lon=-122.3945&lat=37.7925`. The marker sits in the middle of the visible map area above the drawer, not on its edge.
2. Tap the drawer's close button: the marker eases to the vertical center of the full map. Reopen: it returns to the visible half.
3. Search for an address (type three digits, pick a suggestion): the marker lands in the visible half.
4. At desktop width the marker is at the vertical center of the map, as before.

## Clean commits

Execution: Codex (delegated), reviewed and committed by Claude.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/components/__tests__/address-mapper.test.tsx
+++ b/app/components/__tests__/address-mapper.test.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { render, waitFor, screen } from "@testing-library/react";
+import { fireEvent, render, waitFor, screen } from "@testing-library/react";
 import "@testing-library/jest-dom";
 
 import { Provider } from "../ui/provider";
@@ -31,6 +31,7 @@ jest.mock("../map", () => {
     <div
       data-testid="map"
       data-coordinates={JSON.stringify([props.lon, props.lat])}
+      data-bottom-padding-ratio={String(props.bottomPaddingRatio)}
     >
       Mocked Map
     </div>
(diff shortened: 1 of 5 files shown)


danielATiriai:
Folded into #1025 so both commits land in one PR.

PR #1027, opened 2026-09-02
Add landslide hazard layer (Issue #959)

### LINKED ISSUE #959: Add the Landslide dataset to the map
## Context
Add the Landslide dataset to the map:
https://data.sfgov.org/Public-Safety/Landslide-Susceptibility-Hazard-Zones/bna4-itif/about_data

this is a dataset from the DataSF organization on landslide Susceptibility. And it's a verified candidate EDA that we've done to add to our map.

## Definition of Done
* work with our designer and hack-in engineering teams to set up and do the ingestion pipeline and a display of this dataset on site for our next version.

Neelansh-Khare:
## Stacked on #968

This targets `feat/hazard-lookup-composite-api` rather than `develop`, so the diff
shows only the landslide work. #968 needs to merge first; the base can then be
retargeted to `develop`.

Closes #959.

## What this adds

The landslide hazard layer, end to end: a point-lookup endpoint, inclusion in the
composite hazard lookup, ETL for the `landslide_zones` table, and the frontend
wiring to render the layer and report it for a searched address.

## Fixes found while getting the branch green

Four defects, each of which independently prevented the layer from working from
the frontend:

- **Report the most severe overlapping zone.** The point lookup used `.first()`
  with no `ORDER BY`. Zones overlap, so a point inside several returned whichever
  row Postgres scanned first — the seeded test point sits inside gridcodes 3, 8
  and 10 and the endpoint reported 8. For a hazard endpoint that errs in the
  dangerous direction, understating susceptibility at an address, and it is not
  stable across query plans. Now ordered by `gridcode` descending.

- **Serve the router under `/api`.** Every other router uses an `/api` prefix and
  the frontend builds its URLs from `const API_URL = "/api"`, so the frontend was
  requesting `/api/landslide-zones` while the backend served `/landslide-zones`.
  Every landslide request from the UI 404'd; the tests passed only because they
  called the unprefixed path directly.

- **Declare the list route as `""` rather than `"/"`.** The sibling routers all
  use `""`, so `/api/landslide-zones` 404'd while `/api/landslide-zones/` worked.
  The frontend requests the former.

- **Run the landslide ETL.** `init_db` registers `landslide_zones`, so it is
  reported as needing ETL, but `startup.sh` had no case for it and fell through to
  "No ETL mapping for landslide_zones; skipping". The handler existed and was
  never called, so the table stayed empty and the layer had nothing to draw.
  Wiring it in loads 8 zones from data.sfgov.org and generates
  `LandslideZone.geojson`.

## Verification

Against the running stack:

- `backend`: **73 passed**
- `tsc --noEmit`: clean
- frontend: **9 passed, 6 skipped**
- `/api/landslide-zones` and `/api/landslide-zones/is-in-landslide-zone` both 200

## Draft

Opened as a draft pending #968. Mapbox rendering has not been visually verified.


Neelansh-Khare:
@NewtonLC

Neelansh-Khare:
@coderabbitai review

Neelansh-Khare:
@coderabbitai review

